PR #9, opened 2017-03-14
Implement source node set diffusion along a metapath

dhimmel:
Linear algebra by @kkloste


kkloste:
Roger that, done. Thanks!

PR #16, opened 2017-03-18
Refactor tests. Default to numpy.bool_ for adjacency matrix.

dhimmel:
Add `copy=True` option to `diffussion.normalize` to guarantee input matrix is not modified in place.

@kkloste take a look at the modified tests in `test_diffusion.py`. I parametrized them! Hopefully didn't mess them up.


dhimmel:
One benefit of the default `numpy.bool_` encoding is that it reduced the size of `xarray_dataset.pkl.gz` from 18 to 5 megabytes. It will also help with memory issues, but perhaps not a ton, since at some point float matrices will be needed.

PR #21, opened 2017-03-23
Fix dual_normalize testing, closes #19

### LINKED ISSUE #19: Make sure dual_normalize test function tests row *and* column
The dual_normalize test function only checks that column normalization works properly -- correct to check row normalization as well.

https://github.com/greenelab/hetmech/blob/e387eee580cfe3d530a21c92df025afd9f668784/hetmech/test_diffusion.py

kkloste:
The function didn't properly test column normalization, or simultaneous dual normalization, and now it does

### DIFF
--- a/hetmech/test_diffusion.py
+++ b/hetmech/test_diffusion.py
@@ -34,38 +34,37 @@ class TestDualNormalize:
         original = self.get_clean_matrix(dtype)
 
         # Create the matrix expected for single normalization
+        p = exponent # for easier reading
         expect = [
-            [1/3**exponent, 1/3**exponent, 1/3**exponent],
-            [1/2**exponent, 1/2**exponent, 0],
+            [1/3**p, 1/3**p, 1/3**p],
+            [1/2**p, 1/2**p, 0],
             [1, 0, 0],
         ]
         expect = numpy.array(expect, dtype='float64')
 
         # Test row normalization works as expected
         input_matrix = original.copy()
-        matrix = dual_normalize(input_matrix, exponent, 0.0, copy=copy)
+        matrix = dual_normalize(input_matrix, p, 0.0, copy=copy)
         assert numpy.allclose(expect, matrix)
 
         # Test column normalization works as expected
         input_matrix = original.copy()
-        matrix = dual_normalize(input_matrix, 0.0, exponent, copy=copy)
+        matrix = dual_normalize(input_matrix, 0.0, p, copy=copy)
         assert numpy.allclose(numpy.transpose(expect), matrix)
 
         # Create the matrix expected for simultaneous dual normalization
         expect = [
-            [(1/3**exponent) /
-                (1/3**exponent + 1/2**exponent + 1)**exponent,
-             (1/3**exponent) / (1/3**exponent + 1/2**exponent)**exponent,
-             (1/3**exponent) / (1/3**exponent)**exponent],
-            [(1/2**exponent) / 
-                (1/3**exponent + 1/2**exponent + 1)**exponent,
-             (1/2**exponent) /
-                (1/3**exponent + 1/2**exponent)**exponent, 0],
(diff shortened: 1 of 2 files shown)


dhimmel:
@kkloste let's split this into multiple functions, even if that will take some more lines of code. One idea of unit tests is that the function name is specific so when it fails, you know exactly what failed.

So we probably want:

+ `test_dual_normalize_column_damping`
+ `test_dual_normalize_row_damping`
+ `test_dual_normalize_row_and_column_damping`

Perhaps we should even move:

```python
        # Test whether the original matrix is unmodified
        if copy or dtype != 'float64':
            assert numpy.array_equal(original, input_matrix)
        else:
            assert input_matrix is matrix
```

to its own test.

hetmech/test_diffusion.py
```diff
@@ -52,12 +52,16 @@ def test_dual_normalize_column_damping(self, exponent, dtype, copy):
         assert numpy.allclose(numpy.transpose(expect), matrix)
 
         # Create the matrix expected for simultaneous dual normalization
-        expect = [ [(1/3**exponent) / (1/3**exponent + 1/2**exponent + 1)**exponent,
-                    (1/3**exponent) / (1/3**exponent + 1/2**exponent)**exponent,
-                    (1/3**exponent) / (1/3**exponent)**exponent],
-                [(1/2**exponent) / (1/3**exponent + 1/2**exponent + 1)**exponent,
-                 (1/2**exponent) / (1/3**exponent + 1/2**exponent)**exponent, 0],
-                [1 / (1/3**exponent + 1/2**exponent + 1)**exponent, 0, 0],
+        expect = [
```
dhimmel: Perhaps this would be more readable if you prefaced it with `e = exponent`, so it would be shorter?

kkloste: Good idea!

hetmech/test_diffusion.py
```diff
@@ -32,16 +32,39 @@ def test_dual_normalize_passthrough(self, dtype):
...
+             (1/3**p) / (1/3**p + 1/2**p)**p,
+             (1/3**p) / (1/3**p)**p],
+            [(1/2**p) / (1/3**p + 1/2**p + 1)**p,
+             (1/2**p) / (1/3**p + 1/2**p)**p,
+             0],
+            [1 / (1/3**p + 1/2**p + 1)**p, 0, 0],
+        ]
+        expect = numpy.array(expect, dtype='float64')
+        input_matrix = original.copy()
+        matrix = dual_normalize(input_matrix, p, p, copy=copy)
```
dhimmel: When you split this out to it's own function, we should probably test dual exponent with different powers for row and column.

hetmech/test_diffusion.py
```diff
@@ -32,16 +32,39 @@ def test_dual_normalize_passthrough(self, dtype):
     def test_dual_normalize_column_damping(self, exponent, dtype, copy):
         """Test column_damping"""
         original = self.get_clean_matrix(dtype)
-        input_matrix = original.copy()
-        matrix = dual_normalize(input_matrix, exponent, 0.0, copy=copy)
 
-        # Test the normalized matrix is as expected
+        # Create the matrix expected for single normalization
+        p = exponent # for easier reading
```
dhimmel:
```
hetmech/test_diffusion.py:37:21: E261 at least two spaces before inline comment
```

The greedy flake8 wants another space before the `#`

kkloste: hahaha

hetmech/test_diffusion.py
kkloste: Was thinking the same thing.

kkloste:
@dhimmel -- I split the tests into to functions: one for simultaneously using both row and column damping, and the other for individual uses of row and column. I tried splitting into three, but there was so much redundant code, it didn't seem worth it.

From my perspective, this pull request is ready to merge.

PR #25, opened 2017-03-28
[Do not merge] Rename and refactor DWWC functionality

dhimmel:
@kkloste this PR does a few different things (see individual commits).

Check it over with heightened scrutiny on function names and arguments.

### DIFF
--- a/1.diffusion-example.ipynb
+++ b/1.diffusion-example.ipynb
@@ -24,7 +24,7 @@
     "from neo4j.v1 import GraphDatabase\n",
     "import hetio.readwrite\n",
     "\n",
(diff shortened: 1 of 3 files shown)


dhimmel:
The main goal here is to name and declare functions in a modular way to support the multipurpose application of these functions.

Also free to chat today, if you want to go over in more detail.

dhimmel:
What I've called `degree_weight_adjacency_matrix` here is wrong (see https://github.com/greenelab/hetmech/pull/26#issuecomment-289909436). Perhaps we should call that `diffuse_adjacency_matrix`?

Issue #20, opened 2017-03-23
Implement Degree Weighted Path Count (DWPC)


kkloste:
The diffusion code we have in https://github.com/greenelab/hetmech/blob/e387eee580cfe3d530a21c92df025afd9f668784/hetmech/diffusion.py
can be made to compute the DWPC for an input metapath.
Implement this functionality -- @dhimmel in which module should we put this function?

dhimmel:
> which module should we put this function?

I think it could go in `diffusion.py`. Or you can make a new module `path_count.py`.

We will also want a basic `path_count` function that will count the number of paths without any degree normalization.

For reference, here are existing implementations of DWPC:

+ [visual](https://doi.org/10.1371/journal.pcbi.1004259.g002 "Figure 2: Heterogeneous Network Edge Prediction: A Data Integration Approach to Prioritize Disease-Associated Genes")
+ [cypher](https://thinklab.com/discussion/using-the-neo4j-graph-database-for-hetnets/112#6)
+ [python](https://github.com/dhimmel/hetio/blob/488d603dbec663b38a95ba3315d83ae4bb4a5f21/hetio/pathtools.py#L7-L15)

For now, let's ignore the issue of duplicate nodes in a path.

kkloste:
A note for the sake of technical correctness: a path with duplicate nodes is called a "trail", and is not a "path".

kkloste:
The current `diffusion.py` functionality already enables counting the number of walks between any source node and any target node -- simply call `diffusion.diffuse_along_metapath` with `row_damping = 0` and `column_damping = 0`.

dhimmel:
> The current diffusion.py functionality already enables counting the number of walks between any source node and any target node

Cool. `diffusion.diffuse_along_metapath` is very flexible.

> A note for the sake of technical correctness: a path with duplicate nodes is called a "trail", and is not a "path".

So should we call what we're calculating DWWC (degree weighted walk count)?

kkloste:
> So should we call what we're calculating DWWC (degree weighted walk count)?

I think yes, otherwise the name is misleading. It is worth noting, though, that DWWC = DWPC if the metapath does not repeat node-types.

kkloste:
@dhimmel -- given that `diffuse_along_metapath` implements DWPC exactly in the case that no node-types are repeated, do you still want a separate function that implements DWPC in the case that there *are* repeated node-types?

dhimmel:
[...]

PR #26, opened 2017-03-28
[Do not merge] Implement DWWC

dhimmel:
xref https://github.com/greenelab/hetmech/issues/20

Dependent on #25.

@kkloste this PR implements DWWC and evaluates it in a notebook.

The speed is astounding. Easily a several thousand fold improvement over the neo4j path traversal runtime. Currently, the method produces an equal PC but a different DWWC than the cypher implementation. Something to investigate.

### DIFF
--- a/1.diffusion-example.ipynb
+++ b/1.diffusion-example.ipynb
@@ -24,7 +24,7 @@
     "from neo4j.v1 import GraphDatabase\n",
     "import hetio.readwrite\n",
     "\n",
-    "from hetmech.diffusion import diffuse_along_metapath"
+    "from hetmech.diffusion import dwwc_diffusion"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


kkloste:
This technically does not implement degree-weighted-path-count (or even DWWC) as I understand it, because of the way the matrix normalization is executed: when scaling the rows, the code changes what each column will sum to before the column normalization is computed.

The good news is it shouldn't be too hard to make the necessary changes (and won't increase complexity).

dhimmel:
> This technically does not implement degree-weighted-path-count (or even DWWC) as I understand it

Fantastic. Perhaps the correct implementation will fix the discrepancy with the Cypher implementation.

I understand now what I've done... I think.

+ DWWC/DWPC: must take row/column sums before any normalization
+ diffusion: sequentially normalize.

Means that the current function names in #25 are busted.

PR #32, opened 2017-03-31
Transpose adjacency matrices so source nodes are rows

### LINKED ISSUE #27: Are our adjacency matrices transposed?
Currently, our adjacency matrices for metaedges encode source nodes as columns and target nodes as rows. I'm starting to think this is backwards. Two observations:

1. to access a specific value, we must go `matrix[target_pos, source_pos]`. It's more natural to declare source then target.

2. In the proposed [DWWC implemenation](https://github.com/dhimmel/hetmech/blob/2ed28bb589caa909a85dcf979207d4dae827eb7d/hetmech/diffusion.py#L160), we have to do `dwwc_matrix = adj_mat @ dwwc_matrix` rather than `dwwc_matrix = dwwc_matrix @ adj_mat`. The later is preferable in that we could use `functools.reduce`.

@kkloste what do you think?

dhimmel:
Closes #27.

### DIFF
--- a/hetmech/diffusion.py
+++ b/hetmech/diffusion.py
@@ -12,12 +12,15 @@ def diffusion_step(
     """
     Return the diffusion adjacency matrix produced by the input matrix
     with the specified row and column normalization exponents.
+    Note: the row normalization is performed second, so if a value
+    of row_damping=1 is used, the output will be a row-stochastic
+    matrix regardless of choice of column normalization.
 
     Parameters
     ==========
     matrix : numpy.ndarray
         adjacency matrix for a given metaedge, where the source nodes are
-        columns and the target nodes are rows
+        rows and the target nodes are columns
     row_damping : int or float
         exponent to use in scaling each node's row by its in-degree
     column_damping : int or float
@@ -38,25 +41,25 @@ def diffusion_step(
     matrix = numpy.array(matrix, numpy.float64, copy=copy)
     assert matrix.ndim == 2
 
-    # Perform row normalization
-    if row_damping != 0:
-        row_sums = matrix.sum(axis=1)
-        matrix = normalize(matrix, row_sums, 'rows', row_damping)
-
     # Perform column normalization
     if column_damping != 0:
         column_sums = matrix.sum(axis=0)
         matrix = normalize(matrix, column_sums, 'columns', column_damping)
 
+    # Perform row normalization
+    if row_damping != 0:
+        row_sums = matrix.sum(axis=1)
+        matrix = normalize(matrix, row_sums, 'rows', row_damping)
+
     return matrix
 
 
-def diffusion(
+def diffuse(
         graph,
         metapath,
         source_node_weights,
-        column_damping=1,
-        row_damping=0,
+        column_damping=0,
+        row_damping=1,
         ):
     """
     Performs diffusion from the specified source nodes.
@@ -90,7 +93,7 @@ def diffusion(
         adjacency_matrix = diffusion_step(
             adjacency_matrix, row_damping, column_damping)
 
-        node_scores = adjacency_matrix @ node_scores
+        node_scores = node_scores @ adjacency_matrix
 
     target_metanode = metapath.target()
     target_node_to_position = get_node_to_position(graph, target_metanode)
--- a/hetmech/test_diffusion.py
+++ b/hetmech/test_diffusion.py
@@ -58,13 +58,13 @@ class TestDualNormalize:
         # Create the matrix expected for simultaneous dual normalization
         pr, pc = row_damping, column_damping
         expect = [
-            [(1/3**pr) / (1/3**pr + 1/2**pr + 1)**pc,
-             (1/3**pr) / (1/3**pr + 1/2**pr)**pc,
-             (1/3**pr) / (1/3**pr)**pc],
-            [(1/2**pr) / (1/3**pr + 1/2**pr + 1)**pc,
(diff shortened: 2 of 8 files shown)


dhimmel:
@kkloste the transposition I did in `hetmech/diffusion.py` of https://github.com/greenelab/hetmech/pull/32/commits/a5bd2bd62ce1be7f97b0e6be45c956552f31f948 gives different results in the diffusion notebook.

I'm assuming this is because I need to do a bit more than just switch the order of multiplication.

hetmech/diffusion.py
```diff
@@ -90,7 +90,7 @@ def diffusion(
         adjacency_matrix = diffusion_step(
             adjacency_matrix, row_damping, column_damping)
 
-        node_scores = adjacency_matrix @ node_scores
+        node_scores = node_scores @ adjacency_matrix
```
dhimmel: @kkloste do we need node scores to be transposed as well?

kkloste:
@dhimmel  Yes, we have to transpose the normalizations, too.

Suppose a metapath is a sequence of 4 meta edges, and we compute the diffusion by multiplying together the normalized diffusion matrices for each edge type:
P4 * P3 * P2 * P1

where the source nodes correspond to the columns of matrix P1, the target nodes correspond to the rows of matrix P4. In this context, what we've been doing to normalize is scale each row by that node's degree, then normalizing each column to sum to 1 (i.e. be "juice-preserving").

The desired new way of performing this computation is to have rows represent the source nodes, and columns represent target nodes, i.e. take the transpose.

Mathematically, we're doing

`( P4 * P3 * P2 * P1 )^T = P1^T * P2^T * P3^T * P4^T`

In implementation, this could be achieved by taking the transpose of each matrix we had been working with (e.g. P1^T ) and multiplying in the reverse order, e.g.
```
node_scores = P1^T
node_scores = node_scores @ P2^T
node_scores = node_scores @ P3^T
```
instead of
```
node_scores = P1
node_scores = P2 @ node_scores
node_scores = P3 @ node_scores
```

Rather than getting the matrix A1 then normalizing it to be P1 then taking the transpose, the efficient thing is to get the transpose of the matrix A1 from the start, i.e. re-write `metaedge_to_adjacency_matrix` to output the transpose of what it's currently outputting,
then change the way we normalize during `diffusion_step`.

If you can re-write `metaedge_to_adjacency_matrix` to output the transpose of what it's currently producing, I can alter the normalization in diffusion step.

hetmech/diffusion.py
kkloste: Currently `node_scores` is initialized as a row vector, which works for our current purposes.
No need to make a change here.

kkloste:
Uh, I inadvertently started a review. I don't know how that happened. Those were intended to be comments.

kkloste:
This is my first time working on a PR from a fork. I'm a little confused how I should make edits to a file.
I want to make changes to `diffusion.py` -- do I access that file from here: https://github.com/dhimmel/hetmech/blob/transpose/hetmech/diffusion.py   ?
I assume I need to work on the dhimmel version of the file?

dhimmel:
> I want to make changes to diffusion.py -- do I access that file from https://github.com/dhimmel/hetmech/blob/transpose/hetmech/diffusion.py?

@kkloste usually you would create a pull request that requests to modify my `dhimmel:transpose` branch.

But since I have "Allow edits from maintainers." selected, you should be able to click the pencil from the link you mentioned and make the changes using the GitHub UI directly to my branch. Or you can probably make a commit locally and push it to `dhimmel:transpose`.

kkloste:
Ok, cool, thanks.

kkloste:
We have test functions for the normalizing function, I thought?

dhimmel:
> We have test functions for the normalizing function, I thought?

My bad.. yeah. and they're failing as they should

hetmech/test_diffusion.py
```diff
@@ -56,7 +56,7 @@ def test_diffusion_step_row_and_column_damping(
         input_matrix = self.get_clean_matrix()
 
         # Create the matrix expected for simultaneous dual normalization
-        pr, pc = row_damping, column_damping
+        pc, pr = row_damping, column_damping
```
dhimmel: I made these tests pass, by this ugly hack... would it be hard to rewrite `expect` so we can get rid of the transpose below and this change?

hetmech/diffusion.py
dhimmel: > No need to make a change here.

Got it, fixing `diffustion_step` was all that was necessary. That is great!

hetmech/test_diffusion.py
kkloste: Not hard, just tedious :-)
I'll take care of it!

dhimmel:
@kkloste thanks for fixing. I'm going to squash all the diffusion commits, but the final state will be the same

dhimmel:
@kkloste you can merge this once you approve. Don't use squash merge... use "Create a merge commit" since I want to keep these 4 commits separate.

In the commit message body, you can add `Closes #27.`

kkloste:
Hm. For some reason I am unable to merge. I might not have necessary permissions?
All I can do is close, so I'm closing.

dhimmel:
> Hm. For some reason I am unable to merge. I might not have necessary permissions?

Because, it hasn't received an "Approved" review. You have to provide the review first.

kkloste:
Ah! Thanks for the lesson :-)

PR #33, opened 2017-04-01
Test DW and DWPC computation

dhimmel:
Tests DWPC computation from https://doi.org/10.1371/journal.pcbi.1004259.g002.

See panel D of:

### DIFF
--- a/hetmech/test_degree_weight.py
+++ b/hetmech/test_degree_weight.py
@@ -16,14 +16,9 @@ def test_disease_gene_example_dwwc():
     )
     graph = hetio.readwrite.read_graph(url)
     metagraph = graph.metagraph
-    metapath = metagraph.metapath_from_abbrev('GiGaD')
-
-    # Check that metapath does not contain duplicate metanodes,
-    # i.e. the special case applies where DWWC is equivalent to DWPC.
-    metanodes = metapath.get_nodes()
-    assert len(metanodes) == len(set(metanodes))
 
     # Compute GiGaD path count and DWWC matrices
+    metapath = metagraph.metapath_from_abbrev('GiGaD')
     pc_matrix = dwwc(graph, metapath, damping=0)
     dwwc_matrix = dwwc(graph, metapath, damping=0.5)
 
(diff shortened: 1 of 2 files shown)


dhimmel:
@kkloste, we can use the same hetnet to make a test for `diffusion.diffuse`

hetmech/test_degree_weight.py
```diff
@@ -0,0 +1,31 @@
...
+        'test/data/disease-gene-example-graph.json',
+    )
+    graph = hetio.readwrite.read_graph(url)
+    metagraph = graph.metagraph
+
+    # Compute GiGaD path count and DWPC matrices
+    metapath = metagraph.metapath_from_abbrev('GiGaD')
+    pc_matrix = dwwc(graph, metapath, damping=0)
+    dwwc_matrix = dwwc(graph, metapath, damping=0.5)
+
```
kkloste: I just want to clarify that in *this* case `dwwc( graph, metapath, damping=0)` produces the path count, but only because the requested metapath GiGaD does not repeat nodes.

In general this could would return the walk count.

kkloste: Similarly: dwwc_matrix happens to be the exact same as the degree-weighted-path-count matrix in this example, for the same reason: GiGaD doesn't repeat any any nodes.

Not to be pedantic, just want to make sure terminology is clear.

kkloste:
Code all looks correct; I've made comments about terminology just to make sure everything's crystal clear.

hetmech/test_degree_weight.py
dhimmel: https://github.com/greenelab/hetmech/pull/33/commits/2e47b99ab2cc310ffdb8bb2da8dc4d6b9756eda8 makes this explicit in the test

dhimmel:
@kkloste good eye. https://github.com/greenelab/hetmech/pull/33/commits/2e47b99ab2cc310ffdb8bb2da8dc4d6b9756eda8 was a stupid commit. https://github.com/greenelab/hetmech/pull/33/commits/59b685edb07e914966d9da0919cbee7e57e67f62 is adds the documentation to explain your point.

dhimmel:
@kkloste for your reference, this is a PR that should be squash merged.

hetmech/test_degree_weight.py
```diff
@@ -32,5 +27,10 @@ def test_disease_gene_example_dwwc():
     disease_index = get_node_to_position(graph, 'Disease')
     i = gene_index[graph.node_dict['Gene', 'IRF1']]
     j = disease_index[graph.node_dict['Disease', 'Multiple Sclerosis']]
+
+    # Warning: the WC (walk count) and PC (path count) are only equivalent
+    # because none of the GiGaD paths contain duplicate nodes. Since, GiGaD
+    # contains duplicate metanodes, WC and PC are not gauranteed to be the
+    # same. However, they happen to be equivalent for this example.
```
kkloste: I can't tell from this comment whether the metapath GiGaD contains duplicate nodes (line 33 seems to say yes) or not (line 32 seems to say no).

dhimmel: The metapath contains duplicate metanodes (two Gene metanodes).

The actual paths in the example hetnet do not have duplicate nodes.

kkloste: Ah, that clears it all up.

PR #34, opened 2017-04-01
WIP: Implement matrix DWPC

dhimmel:
@kkloste, I created a notebook for you to play with. Currently, it computes the wrong DWPC because it is computing DWWC. The goal will be to implement a matrix method that can get the same results as the Cypher and hetio methods.

### DIFF
--- a/hetmech/path_count.py
+++ b/hetmech/path_count.py
@@ -1,4 +1,3 @@
-import numpy
 import collections
 from scipy.sparse import csr_matrix
 import scipy.sparse
@@ -10,21 +9,22 @@ from .degree_weight import dwwc_step
 def dwpc(graph, metapath, damping=1.0, verbose=False):
     """
     Compute the degree-weighted path count (DWPC).
-    First checks that no more than one metanode type is repeated.
+    First checks that no more than one metanode type is repeated,
+    then separates metapath into three segments: head, loop, tail.
+    The 'loop' segment contains all instances of repeated nodes;
+    the three segments are handled with different multiplication rules.
     """
     print("Calling DWPC")
     dwpc_matrix = None
     # First read through sequence of metanode types
     nodetype_sequence = collections.defaultdict(int)
     for metaedge in metapath:
-        # metanodes = metaedge.get_nodes()
         nodetype_sequence[metaedge.source] += 1
         nodetype_sequence[metaedge.target] += 1
     # Identify the repeated node type
     # and check that no more than 1 metanodetype is repeated
-    repeated_node = filter(lambda x: nodetype_sequence[x] > 2,
-                           nodetype_sequence)
-    repeated_node = list(repeated_node)
+    repeated_node = list(filter(lambda x: nodetype_sequence[x] > 2,
+                         nodetype_sequence))
     if len(repeated_node) > 1:
         if verbose:
             print("Input metapath repeats more than one nodetype")
@@ -38,15 +38,16 @@ def dwpc(graph, metapath, damping=1.0, verbose=False):
             adj_mat = metaedge_to_adjacency_matrix(graph, metaedge)
             adj_mat = dwwc_step(adj_mat, damping, damping)
             if dwpc_matrix is None:
-                dwpc_matrix = csr_matrix(adj_mat)
+                dwpc_matrix = csr_matrix(adj_mat).copy()
             else:
                 dwpc_matrix = dwpc_matrix @ adj_mat
     else:
         # Determine start/endpoints for left,loop,right
         if verbose:
             print("Input metapath repeats exactly one nodetype")
-            print("Repeated node list: {}, type {}".format(repeated_node, repeated_node[0]))
-        
+            print("Repeated node list: {}, type {}"
+                  "".format(repeated_node, repeated_node[0]))
+
         first_appearance = None
         last_appearance = None
         for idx, metaedge in enumerate(metapath):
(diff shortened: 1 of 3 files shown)


hetmech/path_count.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def dwpc(graph, metapath, damping=1.0):
+    """
+    Compute the degree-weighted path count (DWPC).
+    First checks that no more than one metanode type is repeated.
+    """
+    dwpc_matrix = None
+    # First read through sequence of metanode types
+    nodetype_sequence = collections.defaultdict(int)
```
dhimmel: I think this is what you are doing:

```python
metanode_counts = collections.Counter(metapath.get_nodes())
```

hetmech/path_count.py
```diff
@@ -0,0 +1,72 @@
...
+                else:
+                    left_matrix = left_matrix @ adj_mat
+            elif idx < last_appearance:  # i.e. metanodes[0] = repeatednode
+                if dwpc_matrix is None:
+                    dwpc_matrix = adj_mat
+                else:
+                    dwpc_matrix = dwpc_matrix @ adj_mat
+                    # if endpoints are same type, subtract diagonal after mult
+                    if metanodes[1] == repeated_node[0]:
+                        dwpc_matrix -= numpy.diag(numpy.diag(dwpc_matrix))
```
dhimmel: So this is the magic...

kkloste: There is some careful bookkeeping that must be done so it works, but essentially we subtract the diagonal of the matrix product anytime the source and target nodes (of the walk represented by that matrix product) are the same and this removes all intersecting walks, i.e. anything that isn't a path.

But we can only do that careful bookkeeping in an easy manner if no more than a single node type is repeated.

kkloste:
Now that I'm trying to run dwpc and dwwc, I'm getting this error in `hetmech.matrix.py`.
I've look through the implementation of the class `MetaGraph` in `https://github.com/dhimmel/hetio/blob/master/hetio/hetnet.py` and can't find any function that seems similar to `get_metanode_to_nodes()`, so I'm not sure how to fix this...

```
/home/kyle/research/hetmech-dhimmel/hetmech/matrix.py in get_node_to_position(graph, metanode)
     13         # metanode is a name
     14         metanode = graph.metagraph.node_dict[metanode]
---> 15     metanode_to_nodes = graph.metagraph.get_metanode_to_nodes()
     16     nodes = sorted(metanode_to_nodes[metanode])
     17     node_to_position = OrderedDict((n, i) for i, n in enumerate(nodes))

AttributeError: 'MetaGraph' object has no attribute 'get_metanode_to_nodes'
```

kkloste:
@dhimmel  Important question:  the cypher and hetio methods in this notebook compute the dwpc between a single node and one other single node.

In contrast, the method I just implemented currently computes the dwpc between all nodes in the input source-metanode-type and all nodes in the input target-metanode-type.

Do you want this kind of "all-pairs dwpc" function? I'm not sure whether a single-pair-dwpc function would benefit from this matrix approach -- the cypher and hetio methods seem pretty efficient as is...

dhimmel:
> In contrast, the method I just implemented currently computes the dwpc between all nodes in the input source-metanode-type and all nodes in the input target-metanode-type.

This is why the path traversal method is weak and the matrix method is strong! I've found that in most real use cases, we need to know DWPCs for many pairs of nodes, not just one.

In the future, it's likely that we'll care about DWPC from a subset of source nodes to all target nodes or vice versa. But one can always still compute the entire source-target DWPC matrix. Let's continue to do this unless performance becomes a problem.

kkloste:
Ok, @dhimmel , I feel comfortable calling this a working `dwpc` function.
I'm working on better documenting it and writing a more exhaustive test, but if you're curious about performance, I encourage you to try running `dwpc` from `hetmech.path_count` on some metapaths.

It computes all dwpc and all path counts for all the disease/compound pairs in ~ 7 seconds, about 5000 times faster than we'd expect hetio to perform the same computations.

dhimmel:
> 5000 times faster than we'd expect hetio to perform the same computations.

Nice!

Will play around and take a look. If not today, tomorrow.

kkloste:
@dhimmel After my most recent commit (including a test function for the new dwpc), the Travis CI build failed. I had spent a while getting the test function to work on my local machine, so dwpc should have passed the test function. The job log shows the following output for the tests:
```
hetmech/test_degree_weight.py .
hetmech/test_diffusion.py ..........................................
hetmech/test_hetio.py .
hetmech/test_path_count.py /home/travis/.travis/job_stages: line 54:  1885 Killed                  pytest hetmech
```

There is no output about any specific test failing, so I'm not sure what to make of this.
Any advice?

Regardless of this hiccup, as I said before I spent a while writing the test function and getting it to work locally, so I feel more confident saying this implementation of dwpc is ready for use on non-corner-case metapaths.

dhimmel:
> There is no output about any specific test failing, so I'm not sure what to make of this. 

Yeah I've never seen `Killed` before [here](https://travis-ci.org/greenelab/hetmech/builds/235309828#L288). Is it possible the test uses lot's of memory, so Travis killed it? I'll keep investigating. For now don't worry about it.

kkloste:
Gotcha, cool.
Then from my end, DWPC is good to go, until we make more progress getting the hetmech.matrix functions compatible with sparse matrices

hetmech/path_count.py
```diff
@@ -0,0 +1,135 @@
...
+    repeated_node = list(filter(lambda x: nodetype_sequence[x] > 2,
+                         nodetype_sequence))
+    if len(repeated_node) > 1:
+        if verbose:
+            print("Input metapath repeats more than one nodetype")
+        return None
+
+    # Compute DWPC
+    dwpc_matrix = None
+    elif len(repeated_node) == 0:  # case 1: no repeated node types
```
dhimmel: Would it make sense just to call the DWWC method here?

hetmech/path_count.py
```diff
@@ -0,0 +1,135 @@
...
+        metapath for which function computes DWPCs
+    damping : scalar
+        exponent of degree in degree-weighting of DWPC
+    verbose : bool
+        set to True to have function print to screen
+        (temporary, for debugging)
+    """
+
+    # Check sequence of metanode types for repeats
+    nodetype_sequence = collections.defaultdict(int)
```
dhimmel: I think you want `collections.Counter` rather than `collections.defaultdict`.

hetmech/path_count.py
```diff
@@ -0,0 +1,135 @@
...
+    for metaedge in metapath:
+        nodetype_sequence[metaedge.source] += 1
+        nodetype_sequence[metaedge.target] += 1
```
dhimmel: Would it be better to do `metapath.get_nodes()` to get the list of metanodes? Right now you're double counting some metanodes (all intermediary metanodes) but not the terminal metanodes. For example, a `CbGcC` path would only have the C metanode twice, despite having a duplication.

hetmech/path_count.py
```diff
@@ -0,0 +1,135 @@
...
+    # Check that no more than 1 metanodetype is repeated
+    repeated_node = list(filter(lambda x: nodetype_sequence[x] > 2,
+                         nodetype_sequence))
+    if len(repeated_node) > 1:
+        if verbose:
```
dhimmel: Instead use:

```python
raise NotImplementedError("Metapath repeats more than one metanode")
```

hetmech/path_count.py
kkloste: Good point, change made

hetmech/path_count.py
kkloste: Good point; I'm still learning about exceptions and `raise`, thanks!

hetmech/path_count.py
```diff
@@ -0,0 +1,135 @@
...
+            print("Input metapath repeats more than one nodetype")
+        return None
+
+    # Compute DWPC
+    dwpc_matrix = None
```
dhimmel: This dwpc_matrix in between the `if` and `elif` is causing a [syntax error](https://travis-ci.org/greenelab/hetmech/builds/235327270#L304).

hetmech/path_count.py
kkloste: When I originally explored this, after your initial comment, I couldn't get things working --- I just tried again and figured out I wasn't converting the output of `metapath.get_nodes()` to strings properly. Now it's working, and the code looks cleaner for it :-)

kkloste: I had saved that corner case (when source and target nodes were the only repeated nodes) for last, then forgot to come back and handle it. Good catch! Now it's handled.

hetmech/path_count.py
kkloste: Good call.

kkloste:
Finished changing code in light of most recent comments, @dhimmel .
Thanks for the feedback!

hetmech/test_path_count.py
```diff
@@ -0,0 +1,95 @@
...
+
+    assert numpy.allclose(dwpc_matrix[i_idx, :].todense(), cypher_output)
+
+
+def test_all():
+    """Load objects for all other tests to use"""
+    print("Preprocessing begin.")
+    url = 'https://github.com/dhimmel/hetionet/raw/' + \
+          '76550e6c93fbe92124edc71725e8c7dd4ca8b1f5/' + \
+          'hetnet/json/hetionet-v1.0.json.bz2'
```
kkloste: Ah, gotcha.
It'd be great to have a smaller network to test on!

dhimmel:
Summarizing what @kkloste and I discussed today in our 11:11 videochat.

1. The current matrix DWPC algorithm supports a single duplicated metanode. That metanode can be duplicated many times however. The algorithm is easy to expand to multiple metanode duplications as long as the duplicated subpaths are not interweaved (overlapping).
2. We may be able to simplify the current code, but outsourcing all DWWC computation to that function, while having the DWPC function break the metapath into segments and multiply the results of each segment.
3. It is my job to create some hetnets that are ideal for testing matrix DWPC.
4. @kkloste will continue working on his discoveries, namely generalizing the duplication handling and supporting overlapping segments of duplication.

dhimmel:
# Testing subgraph

@kkloste in https://github.com/dhimmel/hetio/commit/c1d47a942f2387cde4ee40c67b7dc5bd93b598f8 I created a subgraph with 3060 nodes and 18474 edges ([notebook](https://github.com/dhimmel/hetio/blob/c1d47a942f2387cde4ee40c67b7dc5bd93b598f8/test/data/bupropion-CbGpPWpGaD-subgraph.ipynb)). This subgraph produces the same DWWC/DWPC for the _CbGpPWpGaD_ metapath between bupropion and nicotine dependence. We can switch to this network for development to speed things up:

```python
url = 'https://github.com/dhimmel/hetionet/raw/{}/{}'.format(
    'c1d47a942f2387cde4ee40c67b7dc5bd93b598f8',
    'test/data/bupropion-CbGpPWpGaD-subgraph.json.xz',
)
graph = hetio.readwrite.read_graph(url)
```

I also included GiG relationships so we can test metapaths like _CbGiGaD_ and the hellish _CbGiGiGpPWpGiGaD_.

dhimmel:
@kkloste, I'm going to continue work here in https://github.com/greenelab/hetmech/pull/34. I decided to open a new PR, since I wanted to leave the commit history here for reference.

Issue #37, opened 2017-05-22
Problem with environment installing neo4j, hetio


kkloste:
I created the environment via conda, as per the instructions in the README, but this failed to install neo4j-driver for some reason:

![screenshot from 2017-05-22 11-46-31](https://cloud.githubusercontent.com/assets/4999436/26317212/bb87cf18-3ee4-11e7-82a5-fd748c87ae3f.png)

so I manually installed neo4j-driver via `pip install neo4j-driver`, but then discovered that `hetio` was not properly installed:
![screenshot from 2017-05-22 11-51-03](https://cloud.githubusercontent.com/assets/4999436/26317326/16ed626e-3ee5-11e7-9732-272c1e70dc5a.png)

These error messages occurred when I was definitely using the `hetmech` environment (i.e. via `source activate hetmech`, as per the README).

dhimmel:
I'm guessing that the conda installation succeeded, but the notebook isn't running the kernel from the `hetmech` environment.

I also had some [kernel problems recently](https://github.com/jupyter/jupyter/issues/245). Can you try:

```sh
conda install --channel=conda-forge nb_conda_kernels
```

Run this inside the `hetmech` environment. Then see if you can choose the right environment in the Jupyter notebook kernel menu. If this fixes the issues, then we can add the command to `environment.yml` like [here](https://github.com/greenelab/scihub/blob/b4e09a5500b7b81cf44dddfe1b9c11955ca218dd/environment.yml#L6).

kkloste:
I was just about to report that something fishy is going on -- apparently hetio did successfully install. I tried installing it inside the environment manually, and this happened:

```
❭ pip install hetio
Requirement already satisfied: hetio in /home/hyunil/.conda/envs/hetmech/lib/python3.6/site-packages
Requirement already satisfied: regex in /home/hyunil/.conda/envs/hetmech/lib/python3.6/site-packages (from hetio)
```

So this seems to be an error with how my jupyter notebook is running?
About to try your suggested code.

kkloste:
Ran
> conda install --channel=conda-forge nb_conda_kernels
as suggested (inside the hetmech environment), then loaded the hetmech kernel in the notebook, and here is the output:

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-871df4ccb045> in <module>()
      2 from neo4j.v1 import GraphDatabase
      3 import hetio.readwrite
[...]

PR #38, opened 2017-05-22
Update conda environment

dhimmel:
Explicitly install tqdm & py2neo.
Explicitly install notebook & nb_conda_kernels.

Closes https://github.com/greenelab/hetmech/issues/37

Upgrade other packages to newer versions.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,15 +1,18 @@
 name: hetmech
 dependencies:
+- conda-forge::nb_conda_kernels=2.1.0
 - flake8=3.3.0
-- scipy=0.19.0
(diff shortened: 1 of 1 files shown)


kkloste:
This removes `regex` from the environment, was that intentional?
I think it might be a dependency...

dhimmel:
> This removes regex from the environment, was that intentional?

Yes, `regex` [now gets](https://github.com/dhimmel/hetio/blob/86c1114a2487cbcbaf05c9128d592de9ad955e49/setup.py#L64-L67) automatically installed by pip since it's in `install_requires` for `hetio`.

Issue #36, opened 2017-04-19
Add sparse matrix option in metaedge_to_adjacency_matrix


dhimmel:
@kkloste meet @zietzm (undergrad at Penn starting to work in the Greene Lab). From now until the summer, I'm planning to deploy @zietzm on several small (potentially random) tasks to build his development skills.

In https://github.com/greenelab/hetmech/issues/13 we discussed strategically using scipy.sparse matrices for efficiency. A good first step will be enhancing [`metaedge_to_adjacency_matrix`](https://github.com/greenelab/hetmech/blob/a0e6a2e59602bbf1de11648bb4ae58a53b43ecad/hetmech/matrix.py#L21-L39) to optionally return a scipy.sparse matrix.

I think this will be a good task for @zietzm. We probably want to allow the user to choose which matrix backend to use, e.g. numpy.array, numpy.matrix (potentially), or any of the [scipy.sparse matrices](https://docs.scipy.org/doc/scipy/reference/sparse.html).

@zietzm would be nice if you add tests to the non-existent `test_matrix.py`.

dhimmel:
@zietzm let's aim to keep backwards comparability (good practice) although if you think a backwards incompatible API change is warranted that is also fine.

dhimmel:
Regarding tests, I'd use the example hetnet [used here](https://github.com/greenelab/hetmech/blob/a0e6a2e59602bbf1de11648bb4ae58a53b43ecad/hetmech/test_degree_weight.py#L9-L18). @zietzm perhaps the best way to start is to write tests for the current `metaedge_to_adjacency_matrix` in an initial pull request. Once that's merged, we can move to scipy.sparse support.

kkloste:
Nice to meet you, @zietzm 

That's a great test-writing tip, thanks Daniel!

kkloste:
Now that I'm back on task in this project, implementing an option in `metaedge_to_adjacency_matrix` is a priority for me -- it should make a lot of the matrix methods we've been exploring, including the current implementation of `dwpc` in `path_county.py`,more efficient.

As a general rule, @dhimmel , now that I've seen the data more up-close I can say that these metaedge adjacency matrices absolutely should be handled using sparse data structures:
with the current implementation of `metaedge_to_adjacency_matrix` using dense matrices, attempting the matrix multiplies in `dwpc` used GBs of RAM, and eventually caused the computer to run out of memory.

[...]

PR #39, opened 2017-05-24
Create tests for metaedge_to_adjacency_matrix

zietzm:
Refs https://github.com/greenelab/hetmech/issues/36

Test defines a modified version of the original metaedge_to_adjacency matrix function that creates a numpy array and then converts it to a sparse matrix. It then checks whether matrix.py has a function that gives an identical sparse matrix. This test will pass if the new function version in matrix.py gives exactly a sparse matrix version of what the function originally gave.


hetmech/test_matrix.py
```diff
@@ -0,0 +1,51 @@
...
+    https://doi.org/10.1371/journal.pcbi.1004259.g002
+    '''
+
+    url = 'https://github.com/dhimmel/hetio/raw/{}/{}'.format(
+        '9dc747b8fc4e23ef3437829ffde4d047f2e1bdde',
+        'test/data/disease-gene-example-graph.json',
+    )
+    graph = hetio.readwrite.read_graph(url)
+
+    def inef_metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_):
```
dhimmel: You want to remove this complex logic from the test. For this initial PR, let's just try to add tests for the existing `metaedge_to_adjacency_matrix` function without adding any new functionality. In other words, you'll manually construct the adjecency matrices corresponding to the example graph. You'll hardcode these matrices.

Note that tests are different than other types of coding. In most coding, you want to use logic and avoid hardcoding. In tests, you want to hardcode the expected result and check that the function output matches.

hetmech/test_matrix.py
```diff
@@ -0,0 +1,28 @@
...
+'''
+
+url = 'https://github.com/dhimmel/hetio/raw/{}/{}'.format(
+    '9dc747b8fc4e23ef3437829ffde4d047f2e1bdde',
+    'test/data/disease-gene-example-graph.json',
+)
+graph = hetio.readwrite.read_graph(url)
+
+
+def test_sparse_matrix():
```
dhimmel: Let's define a single test `test_metaedge_to_adjacency_matrix` that tests both the `GiG` and `GaD` matrixes are correct.

hetmech/test_matrix.py
```diff
@@ -0,0 +1,28 @@
...
+    mat = metaedge_to_adjacency_matrix(graph, 'GiG')
+    assert all([mat[2][0, 4], mat[2][0, 2], mat[2][6, 2], mat[2][2, 0]])
```
dhimmel: This is on the right track: it checks for the boolean value of certain cells. Let's be even more explicit by testing that the matrix object is exactly equal. You'll want to construct an actual test matrix like [here](https://github.com/greenelab/hetmech/blob/df8a1fca2193f375e8c8c79be76ac8ab255699ea/hetmech/test_diffusion.py#L14-L19).

hetmech/test_matrix.py
```diff
@@ -0,0 +1,28 @@
...
+data as in test_degree_weight.py Figure 2D of Himmelstein &
+Baranzini (2015) PLOS Comp Bio.
+https://doi.org/10.1371/journal.pcbi.1004259.g002
+'''
+
+url = 'https://github.com/dhimmel/hetio/raw/{}/{}'.format(
+    '9dc747b8fc4e23ef3437829ffde4d047f2e1bdde',
+    'test/data/disease-gene-example-graph.json',
+)
+graph = hetio.readwrite.read_graph(url)
```
dhimmel: This should go inside of the test function (or in another function that the test calls). As is, this graph will be loaded whenever the module is imported.

hetmech/test_matrix.py
```diff
@@ -1,28 +1,36 @@
...
+    Baranzini (2015) PLOS Comp Bio.
+    https://doi.org/10.1371/journal.pcbi.1004259.g002
+    """
+    url = 'https://github.com/dhimmel/hetio/raw/{}/{}'.format(
+        '9dc747b8fc4e23ef3437829ffde4d047f2e1bdde',
+        'test/data/disease-gene-example-graph.json',
+    )
+    graph = hetio.readwrite.read_graph(url)
 
+    mat_gig = metaedge_to_adjacency_matrix(graph, 'GiG')
```
dhimmel: I would do something like the following:

```python
row_names, column_names, adjacency_matrix = metaedge_to_adjacency_matrix(graph, 'GiG')
```

And let's also test the row/column names.

dhimmel:
Let's also test Disease--localization--Tissue in both orderings: e.g. `DlT` and `TlD`. It will be good to test both orientations. These matrices should be small (2 x 2)

hetmech/test_matrix.py
```diff
@@ -0,0 +1,68 @@
+import hetio.readwrite
+from .matrix import metaedge_to_adjacency_matrix
+import numpy as np
+
+
+def test_metaedge_to_adjacency_matrix():
+    """
+    Test the functionality of metaedge_to_adjacency_matrix in
+    generating sparse matrices vs numpy arrays. Uses same test
```
dhimmel: Let's remove "sparse matrices vs". Then ready to merge.

dhimmel:
@zietzm I didn't realize you had write access. Oops. In the future, I'll merge unless I note otherwise. The main thing here is that I'd like to squash all the commits. Will do that on master now.

zietzm:
Sorry about that! Will do next time.

dhimmel:
@cgreene currently any Greene Lab member gets write on all repos. While potentially convenient, this scope is a bit broad IMO. See for example above. I think it's best to have write access explicitly granted. For example, I don't think I should be able to go around changing other lab members repos that I'm not involved in.

From this repos settings:

> The Greene Laboratory organization has their default repository permission set to write. This means that every member of this organization has write access to this repository, regardless of the team and collaborator access specified below.

dhimmel:
@zietzm I squashed into https://github.com/greenelab/hetmech/commit/b11b5f699c879345261187e4b5a283109d88680a.

kkloste:
Finally had a chance to look over some of this work -- looking good, @zietzm !
This will also make it easier to code sparse-matrix functionality and testing.

PR #40, opened 2017-05-31
WIP: Implement matrix DWPC (second PR)

dhimmel:
Continues on https://github.com/greenelab/hetmech/pull/34

Required: rebase on master after https://github.com/greenelab/hetmech/pull/41 merged

PR #41, opened 2017-05-31
Add row/column names to DWWC output

dhimmel:
To save the user the hassle of having to regenerate the names.

### DIFF
--- a/hetmech/test_degree_weight.py
+++ b/hetmech/test_degree_weight.py
@@ -19,14 +19,16 @@ def test_disease_gene_example_dwwc():
 
     # Compute GiGaD path count and DWWC matrices
     metapath = metagraph.metapath_from_abbrev('GiGaD')
-    pc_matrix = dwwc(graph, metapath, damping=0)
-    dwwc_matrix = dwwc(graph, metapath, damping=0.5)
+    rows, cols, pc_matrix = dwwc(graph, metapath, damping=0)
+    rows, cols, dwwc_matrix = dwwc(graph, metapath, damping=0.5)
+
+    # Check row and column name assignment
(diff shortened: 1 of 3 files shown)


dhimmel:
@zietzm or @kkloste: can we get a review?

dhimmel:
> why we seem to want to return the first set of rows and the last set of columns in dwwc

The resulting matrix has rows corresponding to the first matrix and columns corresponding to the last matrix.

> So if Line 50 and 55 in degree_weight.py were omitted and line 59 just returned "rows, cols, dwwc_matrix", the test passes just as well.

Good observation. This is a special case where the broken implementation returns the correct result. If you do `return rows, cols, dwwc_matrix`, you return `rows` and `cols` from the last iteration of the for loop. In this case, the last matrix is Gene by Disease. Genes happen to be the starting type of the metapath. However, had we done a metapath like `CbGiGaD` the broken method will fail.

hetmech/test_degree_weight.py
```diff
@@ -19,14 +18,16 @@ def test_disease_gene_example_dwwc():
 
     # Compute GiGaD path count and DWWC matrices
```
kkloste: because there is a repeated node in the metapath (G) and because this notebook uses `dwwc` and not `dwpc`, the object computed here (`dwwc(graph, metapath, damping=0.5`) is not a path count but a walk count.

dhimmel: I agree the name should be changed to dwwc. This will happen with https://github.com/greenelab/hetmech/pull/40/files#diff-e2d55ecf6a3953dc596c0bdc7ba2792e

PR #42, opened 2017-06-01
Add sparse matrix functionality in metaedge_to_adjacency_matrix

### LINKED ISSUE #36: Add sparse matrix option in metaedge_to_adjacency_matrix

zietzm:
Closes #36 

Edited test_matrix.py to test equality of sparse matrices. Added conversion in degree_weight.py from sparse to dense.

### DIFF
--- a/hetmech/matrix.py
+++ b/hetmech/matrix.py
@@ -38,12 +38,11 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
             data.append(1)
     adjacency_matrix = sparse.csc_matrix((data, (row, col)), shape=shape,
                                          dtype=dtype)
-    if matrix_type == numpy.array:
+    if matrix_type == numpy.array or matrix_type == numpy.ndarray:
         adjacency_matrix = adjacency_matrix.toarray()
     elif matrix_type == numpy.matrix:
         adjacency_matrix = adjacency_matrix.todense()
-    elif matrix_type != type(adjacency_matrix) and issubclass(matrix_type,
-                                                              sparse.spmatrix):
+    else:
         adjacency_matrix = matrix_type(adjacency_matrix)
     row_names = [node.identifier for node in source_nodes]
     column_names = [node.identifier for node in target_node_to_position]
--- a/hetmech/matrix.py
+++ b/hetmech/matrix.py
@@ -40,9 +40,10 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
                                          dtype=dtype)
     if matrix_type == numpy.array:
         adjacency_matrix = adjacency_matrix.toarray()
-    if matrix_type == numpy.matrix:
+    elif matrix_type == numpy.matrix:
         adjacency_matrix = adjacency_matrix.todense()
-    elif matrix_type != type(adjacency_matrix):
+    elif matrix_type != type(adjacency_matrix) and issubclass(matrix_type,
+                                                              sparse.spmatrix):
         adjacency_matrix = matrix_type(adjacency_matrix)
     row_names = [node.identifier for node in source_nodes]
     column_names = [node.identifier for node in target_node_to_position]
--- a/hetmech/matrix.py
+++ b/hetmech/matrix.py
@@ -34,7 +34,7 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
         for edge in source_node.edges[metaedge]:
             j = target_node_to_position[edge.target]
             adjacency_matrix[i, j] = 1
-    adjacency_matrix = matrix_type(adjacency_matrix)
+    adjacency_matrix = matrix_type(adjacency_matrix, dtype=dtype)
     row_names = [node.identifier for node in source_nodes]
     column_names = [node.identifier for node in target_node_to_position]
     return row_names, column_names, adjacency_matrix
--- a/hetmech/matrix.py
+++ b/hetmech/matrix.py
@@ -37,8 +37,9 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
         for edge in source_node.edges[metaedge]:
             row.append(i)
             col.append(target_node_to_position[edge.target])
-            data.append(dtype(1))
-    adjacency_matrix = matrix_type((data, (row, col)), shape=shape)
+            data.append(1)
+    adjacency_matrix = matrix_type((data, (row, col)), shape=shape,
+                                   dtype=dtype)
     row_names = [node.identifier for node in source_nodes]
     column_names = [node.identifier for node in target_node_to_position]
     return row_names, column_names, adjacency_matrix
--- a/hetmech/matrix.py
+++ b/hetmech/matrix.py
@@ -21,7 +21,7 @@ def get_node_to_position(graph, metanode):
 
 
 def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
-                                 mattype=sparse.csc_matrix):
+                                 matrix_type=sparse.csc_matrix):
     """
     Returns an adjacency matrix where source nodes are rows and target
     nodes are columns.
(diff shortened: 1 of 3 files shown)


zietzm:
@dhimmel Will we want to work entirely with sparse matrices, or will conversion to dense be workable for some steps, for example in [dwwc_step](https://github.com/greenelab/hetmech/blob/master/hetmech/degree_weight.py#L34-L40)? 

[In the PR](https://github.com/greenelab/hetmech/pull/42/files#diff-dcb7103539fb7297ec924e8232f7736eR34), I added a line to convert to dense, but I realize doing this whenever we need to use a matrix defeats the purpose of using sparse matrices.

hetmech/test_matrix.py
```diff
@@ -1,9 +1,45 @@
...
+dlt_cols = ['Leukocyte', 'Lung']
+dlt_adj = sparse.csc_matrix([[0, 0],
+                             [1, 0]])
+tld_rows = ['Leukocyte', 'Lung']
+tld_cols = ["Crohn's Disease", 'Multiple Sclerosis']
+tld_adj = sparse.csc_matrix([[0, 1],
+                             [0, 0]])
+
+
+@pytest.mark.parametrize("test_edge,exp_row,exp_col,exp_adj", [
```
dhimmel: We want to parametrize the matrix type (and probably also dtype).

```python
@pytest.mark.parametrize('matrix_type', [numpy.ndarray, scipy.sparse.xxxx, scipy.sparse.xxxx])
```

hetmech/matrix.py
```diff
@@ -34,6 +34,7 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_):
         for edge in source_node.edges[metaedge]:
             j = target_node_to_position[edge.target]
             adjacency_matrix[i, j] = 1
+    adjacency_matrix = matrix_type(adjacency_matrix, dtype=dtype)
```
dhimmel: It would be ideal to avoid `numpy.zeros` in the sparse case. Doing the conversion after creation still requires the full memory footprint prior to conversion.

zietzm:
Note, to test the equality of the two arrays, [I used](https://github.com/greenelab/hetmech/pull/42/commits/987fb562eeb41a61f67b38d269704c840b0fd0cb#diff-1032de08c16632378c6395b14ca65155R70) a sparse matrix and its corresponding `.nnz` function to get the most foolproof method of comparison between two arrays, no matter the different types. 

It was tricky trying to compare two arrays of ambiguous type, and this works, despite the fact that it is not an obvious method. 

This should be a pretty efficient way of checking, I think, because using `==` to compare two sparse matrices gives an efficiency warning which recommends using `!=` as an efficient comparision.

hetmech/test_matrix.py
```diff
@@ -66,9 +62,13 @@ def test_metaedge_to_adjacency_matrix(test_edge, exp_row, exp_col, exp_adj,
...
                                      dtype=dtype)
+    exp_row, exp_col, exp_adj = get_arrays(test_edge, mat_type, dtype)
 
-    assert numpy.array_equal(row_names, exp_row)
-    assert numpy.array_equal(col_names, exp_col)
-    assert not type((exp_adj != adj_mat)).max(exp_adj != adj_mat)
+    assert row_names == exp_row
+    assert col_names == exp_col
+    assert type(adj_mat) == type(exp_adj)  # Arrays same type
+    assert not sparse.csc_matrix(adj_mat != exp_adj).nnz  # Arrays equal
```
zietzm: `adj_mat != exp_adj` gives an array with the element-wise comparison of `adj_mat` and `exp_adj`. If all values are `False`, then the arrays are equal. The problem with using `count_nonzero` is that the matrix type of the comparison array changes depending on the types of matrices under comparison. `sparse.csc_matrix.nnz` is a simply way to check equality, because it gives the number not equal regardless of the types of arrays under review.

hetmech/test_matrix.py
```diff
@@ -1,9 +1,53 @@
...
+import pytest
+from scipy import sparse
+
 from .matrix import metaedge_to_adjacency_matrix
-import numpy as np
 
 
-def test_metaedge_to_adjacency_matrix():
+def get_arrays(edge, mat_type, dtype):  # Define a dictionary
+    rows_dict = {
```
dhimmel: `rows_dict` and `cols_dict` are the same.

hetmech/test_matrix.py
```diff
@@ -1,9 +1,53 @@
...
+                [0, 0]]
+    }
+    return rows_dict[edge], cols_dict[edge], mat_type(adj_dict[edge],
+                                                      dtype=dtype)
+
+
+@pytest.mark.parametrize("test_edge", ['GiG', 'GaD', 'DlT', 'TlD'])
+@pytest.mark.parametrize("mat_type", [numpy.array, sparse.csc_matrix,
+                                      sparse.csr_matrix, numpy.matrix])
+@pytest.mark.parametrize("dtype", [numpy.bool_, numpy.int64, numpy.int32])
```
dhimmel: Let's try a `float64` rather than `int32`.

hetmech/test_matrix.py
```diff
@@ -15,53 +59,16 @@ def test_metaedge_to_adjacency_matrix():
...
-
-    row_names, col_names, adj_mat = metaedge_to_adjacency_matrix(graph, "TlD")
-    assert np.array_equal(row_names, tld_rows)
-    assert np.array_equal(col_names, tld_cols)
-    assert np.array_equal(adj_mat, [[0, 1],
-                                    [0, 0]])
+    assert row_names == exp_row
+    assert col_names == exp_col
+    assert type(adj_mat) == type(exp_adj)  # Arrays same type
+    assert not sparse.csc_matrix(adj_mat != exp_adj).nnz  # Arrays equal
```
dhimmel: Why can't we do `assert adj_mat == exp_adj`?

zietzm: When I do that, I get back `ValueError: The truth value of an array with more than one element is ambiguous.`

hetmech/test_matrix.py
zietzm: Not certain I understand. They contain the same **set** of values, but `rows_dict['GaD']` != `cols_dict['GaD']`.

hetmech/test_matrix.py
dhimmel: From [`numpy.array_equal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.array_equal.html), "True if two arrays have the same shape and elements, False otherwise."

So let's do the same

```
assert adj_mat.shape == exp_adj.shape
assert (adj_mat == exp_adj).all()
```

Assuming this works

hetmech/test_matrix.py
```diff
@@ -1,9 +1,46 @@
...
+                [0, 1],
+                [0, 0],
+                [1, 1],
+                [0, 0]],
+        'DlT': [[0, 0],
+                [1, 0]],
+        'TlD': [[0, 1],
+                [0, 0]]
+    }
+    return node_dict[edge[0]], node_dict[edge[-1]], mat_type(adj_dict[edge],
```
dhimmel: for readability let's do:

```python
row_names = node_dict[edge[0]]
col_names = node_dict[edge[-1]]
adj_matrix = mat_type(adj_dict[edge], dtype=dtype)
return row_names, col_names, adj_matrix
```

This way we improve readability while making the code document itself.

hetmech/test_matrix.py
kkloste: This was interesting to learn -- nice find @zietzm

Issue #43, opened 2017-06-07
Introduce functionality for sparse matrices in diffusion and degree_weight


zietzm:
Subissue of #13 
Add support for scipy.sparse matrices in diffusion.py and degree_weight.py. Matrix type will for now be selected as a function parameter.
`matrix.normalize` can now output a sparse matrix, meaning that the dense intermediates should only be of dimension 1. Support allowing all matrices to be sparse

dhimmel:
Something to keep in mind would be an `auto` mode where the matrices would be converted to dense arrays only once they exceed a density threshold.

zietzm:
Had planned to implement something like that with a later issue and a new PR.

zietzm:
Update:
A method called ```sparse_threshold``` has been implemented in #44 within [degree_weight.dwpc_duplicated_metanode](https://github.com/greenelab/hetmech/pull/44/files#diff-dcb7103539fb7297ec924e8232f7736eR120) and [matrix.metaedge_to_adjacency_matrix](https://github.com/greenelab/hetmech/pull/44/files#diff-2e95c8eb7e7141bd64ce5096a28d34b6R21). A user chooses the threshold above which the matrices should be dense, and conversion occurs whenever the matrix crosses the threshold. 

This method is implemented in two files so that (1.) new matrices will be created that are sparse or dense according to the threshold and (2.) as matrices are multiplied in a row and the density changes, the matrix will also be changed between sparse and dense according to the threshold.

Issue #13, opened 2017-03-16
Switch to sparse matrices


dhimmel:
@kkloste's calculations suggest that our large matrices are sparse.

Switching to a sparse encoding could reduce the Gene by Biological Process matrix from ~1000 MB in RAM to ~15 MB in RAM.

kkloste:
I just tried running our diffusion-example.ipynb notebook on my laptop and it ran out of memory in our diffuse_along_metapath() function, because of these dense matrices.
Further evidence we should look into using sparse matrices.

```
---> 45     adjacency_matrix = numpy.zeros((len(target_node_to_position), len(source_nodes)))
     46     for j, source_node in enumerate(source_nodes):
     47         for edge in source_node.edges[metaedge]:

MemoryError: 
```

dhimmel:
@kkloste how do you choose which [sparse matrix format](https://docs.scipy.org/doc/scipy-0.18.1/reference/sparse.html) to use?

kkloste:
Depends on what kinds of operation(s) account for the bulk of the work you'll perform with the matrix.
We are doing mostly mat-vecs and matrix multiplies, so CSC and CSR offer the most compact storage and most efficient operations. In contrast, if we were instead going to be doing a lot of querying individual edges, then DOK or COO would be more efficient.

I recommend CSC. CSC and CSR offer effectively the same kind of advantages, but our operations (matrix times column vector) should be more efficient in CSC (a column-oriented storage scheme).

kkloste:
I should clarify:

the matrices that we perform the multplications with would be best stored as CSC.
BUT, converting to CSC takes a non-trivial amount of resources -- the gain we get from speeding up multiplications by using sparse matrices is lost if we have to convert a new matrix to CSC each time we perform a multiplication.

That said, if we expect that we can cache a CSC matrix and use it multiple times in the future, the conversion can be worth it if we re-use enough.

kkloste:
The sparse matrix format for the larger hetnet object, I am not sure about.
I have to better understand all the different kinds of data it's holding.
 We should set up a quick skype meeting to discuss this.

kkloste:
It seems that multiplying a numpy dense matrix by a sparse matrix will yield a dense matrix if performed this way:

https://docs.scipy.org/doc/scipy/reference/sparse.html

[...]

PR #44, opened 2017-06-07
Add sparse functionality in degree_weight and diffusion

zietzm:
Closes https://github.com/greenelab/hetmech/issues/13
Closes https://github.com/greenelab/hetmech/issues/43

Updated method for row and column sums to ensure that both numpy.ndarrays and scipy.sparse matrices sums over axes 0 and 1 return a dimension 1 numpy.ndarray with the respective row/column sums. 

Added a function parameter to set a threshold above whose density the matrix should be converted to dense.

diffusion node_scores would be a dense, 1 dimensional array, which works with sparse or dense multiplication

### DIFF
--- a/hetmech/degree_weight.py
+++ b/hetmech/degree_weight.py
@@ -8,15 +8,14 @@ from scipy import sparse
 from .matrix import normalize, metaedge_to_adjacency_matrix
 
 
-def dwwc_step(
-        matrix, row_damping=0, column_damping=0, copy=True):
+def dwwc_step(matrix, row_damping=0, column_damping=0, copy=True):
     """
     Return the degree-weighted adjacency matrix produced by the input matrix
     with the specified row and column normalization exponents.
 
     Parameters
     ==========
-    matrix : numpy.ndarray
+    matrix : numpy.ndarray or scipy.sparse
         adjacency matrix for a given metaedge, where the source nodes are
         rows and the target nodes are columns
     row_damping : int or float
@@ -32,7 +31,7 @@ def dwwc_step(
 
     Returns
     =======
-    numpy.ndarray
+    numpy.ndarray or scipy.sparse
         Normalized matrix with dtype.float64.
     """
     # returns a newly allocated array
@@ -52,14 +51,16 @@ def dwwc_step(
 
 def dwwc(graph, metapath, damping=0.5, auto=False, mat_type=numpy.ndarray):
     """
-    Compute the degree-weighted walk count (DWWC).
+    Compute the degree-weighted walk count (DWWC). Special case of function
+    dwpc_duplicated_metanode.
     """
-    return dwpc_duplicated_metanode(graph, metapath, None, damping)
+    return dwpc_duplicated_metanode(graph, metapath, None, damping, auto=auto,
+                                    mat_type=mat_type)
 
 
 def pairwise(iterable):
     """
-    Yield consequitive pairs of items from the iterable, but skip pairs where
+    Yield consecutive pairs of items from the iterable, but skip pairs where
     the items are equal.
     Modified from recipe in itertools docs at
     https://docs.python.org/3/library/itertools.html
@@ -116,7 +117,8 @@ def get_segments(metagraph, metapath):
     return segments, duplicates
 
 
-def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5):
+def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5,
+                             auto=False, mat_type=numpy.ndarray):
     """
     Compute the degree-weighted path count (DWPC) when a single metanode is
     duplicated (any number of times). User must specify the duplicated
@@ -136,11 +138,14 @@ def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5):
             dwpc_matrix = adj_mat
         else:
             dwpc_matrix = dwpc_matrix @ adj_mat
+            if auto and sparse.issparse(dwpc_matrix):
+                if dwpc_matrix.nnz > (1/3) * numpy.prod(dwpc_matrix.shape):
+                    dwpc_matrix = dwpc_matrix.toarray()
         if metaedge.target == duplicate:
-            # # scipy.sparse method threw ValueError:
-            # # Different number of diagonals and offsets.
-            # diag_matrix = scipy.sparse.diags(dwpc_matrix.diagonal())
-            diag_matrix = numpy.diag(dwpc_matrix.diagonal())
+            if sparse.issparse(dwpc_matrix):
+                diag_matrix = sparse.diags(dwpc_matrix.diagonal())
+            else:
+                diag_matrix = numpy.diag(dwpc_matrix.diagonal())
             dwpc_matrix -= diag_matrix
     return row_names, cols, dwpc_matrix
 
--- a/hetmech/degree_weight.py
+++ b/hetmech/degree_weight.py
@@ -49,15 +49,13 @@ def dwwc_step(matrix, row_damping=0, column_damping=0, copy=True):
     return matrix
 
 
-def dwwc(graph, metapath, damping=0.5, sparse_threshold=0,
-         mat_type=numpy.ndarray):
+def dwwc(graph, metapath, damping=0.5, sparse_threshold=0):
     """
     Compute the degree-weighted walk count (DWWC). Special case of function
     dwpc_duplicated_metanode.
     """
     return dwpc_duplicated_metanode(graph, metapath, None, damping,
-                                    sparse_threshold=sparse_threshold,
-                                    mat_type=mat_type)
+                                    sparse_threshold=sparse_threshold)
 
 
 def pairwise(iterable):
@@ -120,7 +118,7 @@ def get_segments(metagraph, metapath):
 
 
 def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5,
-                             mat_type=numpy.ndarray, sparse_threshold=0):
+                             sparse_threshold=0):
     """
     Compute the degree-weighted path count (DWPC) when a single metanode is
     duplicated (any number of times). User must specify the duplicated
@@ -132,13 +130,9 @@ def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5,
     metapath : hetio.hetnet.MetaPath
     duplicate : hetio.hetnet.MetaNode or None
     damping : float
-    mat_type : type or None
-        type of matrix to be used. Only relevant above the
-        sparse_threshold.
     sparse_threshold : float (0 < sparse_threshold < 1)
         sets the density threshold above which a sparse matrix will be
-        converted to a dense automatically. Supersedes mat_type if the
-        density is below the threshold.
+        converted to a dense automatically.
     """
     if duplicate is not None:
         assert metapath.source() == duplicate
@@ -146,7 +140,7 @@ def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5,
     row_names = None
     for metaedge in metapath:
         rows, cols, adj_mat = \
-            metaedge_to_adjacency_matrix(graph, metaedge, matrix_type=mat_type,
+            metaedge_to_adjacency_matrix(graph, metaedge,
                                          sparse_threshold=sparse_threshold)
         adj_mat = dwwc_step(adj_mat, damping, damping)
         if dwpc_matrix is None:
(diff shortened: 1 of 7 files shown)


hetmech/degree_weight.py
```diff
@@ -30,25 +30,29 @@ def dwwc_step(
...
+    # returns a newly allocated array
+    mat_type = type(matrix)
+    if mat_type == numpy.ndarray:
+        mat_type = numpy.array
+    matrix = mat_type(matrix, dtype=numpy.float64, copy=copy)
     assert matrix.ndim == 2
 
-    row_sums = matrix.sum(axis=1)
-    column_sums = matrix.sum(axis=0)
+    row_sums = numpy.array(matrix.sum(axis=1)).flatten()
```
zietzm: I put that in there because, as far as I can tell, giving `matrix.sum(axis=` type sums gives the same results for sparse and dense. The only difference is that for sums over sparses, the result is given as a list of lists instead of a simple list. The elements are in the right order, I just use `flatten` to make the output identical to `numpy.array` sums. The line in question does not change a dense array's output, and it just gives the same output as it did before.

hetmech/matrix.py
```diff
@@ -38,6 +38,11 @@ def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
             data.append(1)
     adjacency_matrix = sparse.csc_matrix((data, (row, col)), shape=shape,
                                          dtype=dtype)
+    if auto:
```
zietzm: Makes sense. So should it be build into the `dwwc` and `diffusion` functions themselves instead of into the `metaedge_to_adjacency_matrix`?

hetmech/matrix.py
```diff
@@ -69,5 +76,8 @@ def normalize(matrix, vector, axis, damping_exponent):
         vector **= -damping_exponent
     vector[numpy.isinf(vector)] = 0
     shape = (len(vector), 1) if axis == 'rows' else (1, len(vector))
-    matrix *= vector.reshape(shape)
+    if sparse.issparse(matrix):
+        matrix = matrix.multiply(vector.reshape(shape))
```
zietzm: `.multiply` is only a method for sparse matrices. The two methods in the `if` [...] `else` give the same resulting matrices, but the type of each will be the same as the type of input matrix.

hetmech/degree_weight.py
```diff
@@ -125,18 +129,23 @@ def dwpc_duplicated_metanode(graph, metapath, duplicate=None, damping=0.5):
...
             dwpc_matrix = dwpc_matrix @ adj_mat
+            if auto and sparse.issparse(dwpc_matrix):
+                if dwpc_matrix.nnz > (1/3) * numpy.prod(dwpc_matrix.shape):
+                    dwpc_matrix = dwpc_matrix.toarray()
         if metaedge.target == duplicate:
-            # # scipy.sparse method threw ValueError:
-            # # Different number of diagonals and offsets.
-            # diag_matrix = scipy.sparse.diags(dwpc_matrix.diagonal())
-            diag_matrix = numpy.diag(dwpc_matrix.diagonal())
+            if sparse.issparse(dwpc_matrix):
```
zietzm: `(numpy.diag(mat.diagonal()) != sparse.diags(sparse.csc_matrix(mat).diagonal()).toarray()).sum()` = 0
`(mat - numpy.diag(mat.diagonal()) != sparse.csc_matrix(mat) - sparse.diags(sparse.csc_matrix(mat).diagonal())).sum()` = 0
for a created matrix `mat`. This shows that we get the same results for both of these methods.

zietzm: Just noticed that the above added method only works for square matrices. I will add functionality for non-square matrices.

hetmech/diffusion.py
```diff
@@ -37,18 +37,21 @@ def diffusion_step(
     numpy.ndarray
         Normalized matrix with dtype.float64.
     """
-    # returns a newly allocated numpy.ndarray
-    matrix = numpy.array(matrix, numpy.float64, copy=copy)
+    # returns a newly allocated array
+    mat_type = type(matrix)
```
dhimmel: These few lines appear in multiple places... should they be pulled out into a function? We want to minimize the type logic in these functions for the sake of readability.

hetmech/matrix.py
```diff
@@ -19,7 +19,7 @@ def get_node_to_position(graph, metanode):
 
 
 def metaedge_to_adjacency_matrix(graph, metaedge, dtype=numpy.bool_,
-                                 matrix_type=numpy.array):
+                                 auto=False, matrix_type=numpy.ndarray):
```
dhimmel: Can we change `auto` to `sparse_threshold` where `if density > sparse_threshold: convert_to_dense`? Then you can set `sparse_threshold` to 0 to disable the conversion.

hetmech/matrix.py
dhimmel: `vector.reshape(shape)` repeated... just do the reshaped to vector outside of if statement?

hetmech/diffusion.py
zietzm: Sure thing! I'll add it to matrix.py

hetmech/matrix.py
zietzm: Done! Added in latest commits.

zietzm:
@dhimmel I think I have made all the changes that need to be made for the time being. This adds a number of tests, but the total runtime on my machine is ~30 seconds for all tests. (~12 on Travis)

Are there any other  features, edits, or tests that ought to be added to this PR?

hetmech/matrix.py
```diff
@@ -69,5 +75,43 @@ def normalize(matrix, vector, axis, damping_exponent):
...
+    above_thresh = (matrix != 0).sum() / numpy.prod(matrix.shape) >= threshold
+    if sparse.issparse(matrix) and above_thresh:
+        matrix = matrix.toarray()
+    elif not sparse.issparse(matrix) and not above_thresh:
+        matrix = sparse.csc_matrix(matrix)
+    return matrix
+
+
+def new_array(matrix, copy=True):
+    """Returns a newly allocated array"""
```
dhimmel: When `copy=False`, I don't think the array is "newly allocated".

hetmech/matrix.py
```diff
@@ -69,5 +75,43 @@ def normalize(matrix, vector, axis, damping_exponent):
...
+    """
+    above_thresh = (matrix != 0).sum() / numpy.prod(matrix.shape) >= threshold
+    if sparse.issparse(matrix) and above_thresh:
+        matrix = matrix.toarray()
+    elif not sparse.issparse(matrix) and not above_thresh:
+        matrix = sparse.csc_matrix(matrix)
+    return matrix
+
+
+def new_array(matrix, copy=True):
```
dhimmel: Let's call this `copy_array`

hetmech/degree_weight.py
```diff
@@ -114,34 +117,51 @@ def get_segments(metagraph, metapath):
...
     duplicated (any number of times). User must specify the duplicated
     metanode.
+
+    Parameters
+    ==========
+    graph : hetio.hetnet.Graph
+    metapath : hetio.hetnet.MetaPath
+    duplicate : hetio.hetnet.MetaNode or None
+    damping : float
+    sparse_threshold : float (0 < sparse_threshold < 1)
```
dhimmel: Less than or equal

hetmech/degree_weight.py
```diff
@@ -114,34 +117,51 @@ def get_segments(metagraph, metapath):
...
-            # # Different number of diagonals and offsets.
-            # diag_matrix = scipy.sparse.diags(dwpc_matrix.diagonal())
-            diag_matrix = numpy.diag(dwpc_matrix.diagonal())
-            dwpc_matrix -= diag_matrix
+            mat_type = type(dwpc_matrix)
+            if mat_type == numpy.ndarray:
+                mat_type = numpy.array
+            diag_matrix = sparse.diags(dwpc_matrix.diagonal())
+            dwpc_matrix = mat_type(dwpc_matrix - diag_matrix,
+                                   dtype=numpy.float64)
```
dhimmel: Shouldn't `copy = False` here to preserve memory

hetmech/matrix.py
```diff
@@ -69,5 +75,43 @@ def normalize(matrix, vector, axis, damping_exponent):
...
+    threshold : float (0 < threshold < 1)
+        percent nonzero above which the matrix is converted to dense
+
+    Returns
+    =======
+    matrix : numpy.ndarray or scipy.sparse
+    """
+    above_thresh = (matrix != 0).sum() / numpy.prod(matrix.shape) >= threshold
+    if sparse.issparse(matrix) and above_thresh:
+        matrix = matrix.toarray()
```
dhimmel: `return matrix.toarray()`. Then no need elif.

hetmech/test_degree_weight.py
```diff
@@ -31,7 +33,117 @@ def test_disease_gene_example_dwwc():
...
     # because none of the GiGaD paths contain duplicate nodes. Since, GiGaD
-    # contains duplicate metanodes, WC and PC are not gauranteed to be the
+    # contains duplicate metanodes, WC and PC are not guaranteed to be the
     # same. However, they happen to be equivalent for this example.
     assert wc_matrix[i, j] == 3
-    assert dwwc_matrix[i, j] == pytest.approx(0.25 + 0.25 + 32**-0.5)
+    assert dwwc_matrix[i, j] == pytest.approx(0.25 + 0.25 + 32 ** -0.5)
+
+
+def get_expect(m_path, threshold):
```
dhimmel: get_expected, metapath

hetmech/test_degree_weight.py
```diff
@@ -31,7 +33,117 @@ def test_disease_gene_example_dwwc():
...
+    else:
+        assert sparse.issparse(dwwc_mat)
+
+    if exp_type_dwpc == numpy.array:
+        assert isinstance(dwpc_mat, numpy.ndarray)
+    else:
+        assert sparse.issparse(dwpc_mat)
+
+    # Test matrix, row, and column label output
+    assert (dwwc_mat - exp_dwwc).sum() < 0.000001  # Assert approx equal
```
dhimmel: pytest.approx

hetmech/test_matrix.py
```diff
@@ -54,14 +59,13 @@ def test_metaedge_to_adjacency_matrix(test_edge, mat_type, dtype):
...
+    row_names, col_names, adj_mat = metaedge_to_adjacency_matrix(
+            graph, test_edge, dtype=dtype, sparse_threshold=threshold)
+    exp_row, exp_col, exp_adj = get_arrays(test_edge, dtype, threshold)
 
     assert row_names == exp_row
     assert col_names == exp_col
     assert type(adj_mat) == type(exp_adj)
     assert adj_mat.dtype == dtype
     assert adj_mat.shape == exp_adj.shape
-    assert (adj_mat != exp_adj).sum() == 0
```
dhimmel: Explicit not implicit

hetmech/test_path_count.py
```diff
@@ -60,7 +59,7 @@ def test_CbGpPWpGaD_traversal():
     """
     graph = get_bupropion_subgraph()
     compound = 'DB01156'  # Bupropion
-    disease = 'DOID:0050742'  # nicotine dependences
+    disease = 'DOID:0050742'  # nicotine dependencies
```
dhimmel: "nicotine dependence"

hetmech/matrix.py
zietzm: I think I will simplify the `elif`, but we will miss the case where the array is above the threshold and dense (where we want to do nothing at all) if we just put [`else`] `return sparse.csc_matrix(matrix)`. That would be a conversion to sparse that we don't want to do.

hetmech/test_degree_weight.py
```diff
@@ -31,7 +33,117 @@ def test_disease_gene_example_dwwc():
...
+        type_dwpc = type_dict[mat_dict[m_path][1] >= threshold]
+
+    dwwc_mat = type_dwwc(dwwc_mat)
+    dwpc_mat = type_dwpc(dwpc_mat)
+
+    return type_dwwc, dwwc_mat, type_dwpc, dwpc_mat, row_names, col_names
+
+
+@pytest.mark.parametrize('thresh', (0, 0.25, 0.3, 0.5, 0.7, 1))
+@pytest.mark.parametrize('m_path', ('GiGaD', 'GaDaG', 'GeTlD', 'GiG'))
```
zietzm: Will leave `m_path` here to distinguish the metapath as a string (`m_path`) from the metapath object which will be-- in the test function-- `metapath`.

hetmech/matrix.py
dhimmel: You can go if, if, and then a final return matrix in case neither was triggered. The point is elif is equivalent to if when the previous if clause returns.

Issue #47, opened 2017-06-26
DWPC implementation coverage of the Project Rephetio metapaths 


dhimmel:
We currently have implemented the DWPC (see #45) for certain metapaths. The question is, of the 1,206 metapaths in Project Rephetio, how many will our current implementation succeed on?

You can get the metapath info [here](https://github.com/dhimmel/learn/blob/25093893ed53730ab5cfdac49561c4b6bd3376c5/all-features/data/metapaths.json) or [here](https://github.com/dhimmel/learn/blob/25093893ed53730ab5cfdac49561c4b6bd3376c5/prediction/features/metapaths.tsv). An interactive metapath browser for Project Rephetio is also available at http://het.io/repurpose/metapaths.html

zietzm:
**454 / 1206** metapaths will throw an error when running `dwpc`. This error stems from `get_segments` in which an error is raised if there are overlapping segments of duplicate metanodes.

I am currently running the full `dwpc` function for all the 752 metapaths between all compounds and diseases that will run in our current implementation. From this, I hope to compare predictions to Rephetio's predictions, at least for this subset of metapaths. I think it would be really interesting to see where the metapaths we don't support fall in the ranking of metapaths by AUROC and coef. It may be useful to know if the paths we don't support with hetmech are informative or not.

dhimmel:
> 454 / 1206 metapaths will throw an error when running dwpc

@zietzm can you submit a pull request with this notebook?

zietzm:
Sure thing! It is running something at the moment, but I will pull request as soon as it is complete.

kkloste:
I've made some progress this week on the "two repeated metanodes, interleaved" case this week.
Specifically the "ABAB" sub-case ("BAAB" requires a different solution, I believe).
Hopefully this will bring down that "454" to a lower number.

zietzm:
![adjacency_matrix_size](https://user-images.githubusercontent.com/7230303/27748841-f833d398-5d9e-11e7-9680-094648b0f9f4.png)

I wanted to reference the relative sizes here, so I put the table in above. I will hopefully have one soon that will show the time of multiplication for all of the relevant metaedge combinations.

zietzm:
Another random note:
[...]

PR #49, opened 2017-06-28
Replicate Rephetio DWPC values

zietzm:
Refs #47 

Added a file to test the ability of hetmech to replicate Rephetio results. 
~~This preliminary notebook does not have output from the longer DWPC run over all the compatible metapaths, as that will have to be run overnight.~~

Edit:
The information from the run over all compatible metapaths has been added in [this commit](https://github.com/greenelab/hetmech/pull/49/commits/f84fb6e67b4458ccff2618104b388e0643a0e05c)


data/.gitignore
```diff
@@ -1 +1,3 @@
 xarray_dataset.pkl.gz
+hetionet-v1.0.json
```
dhimmel: Can we just load these from versioned URLs instead?

hetmech/rephetio.ipynb
```diff
@@ -0,0 +1,899 @@
...
+   "source": [
+    "# %load matrix.py\n",
+    "from collections import OrderedDict\n",
+    "\n",
+    "import hetio.hetnet\n",
+    "import numpy\n",
+    "from scipy import sparse\n",
+    "\n",
+    "\n",
+    "def get_node_to_position(graph, metanode):\n",
```
dhimmel: Why repeating code that could be imported?

zietzm:
Apologies for the verbose output in the notebook. I had tried running this yesterday but it took too long, and then returned an error. The latest commit includes the information for all 752 as-of-now compatible metapaths. 

One interesting thing I noticed about the output is how clustered the times were around a few different average times. I'm sure that there is a deeper explanation behind this, but I thought I'd just put the histogram of times here on the off chance that it could be useful in the future.

dwpc_time_counts

kkloste:
Neat! I suspect the different clusters are because of different lengths of metapath, and the size of the dimensions of the matrices underlying the metapaths.

@zietzm if you plot runtime vs length of metapath, or vs "maximum dimension matrix involved in metapath" I bet we'd see a pattern emerge.

I am slightly worried about the cluster near 120 seconds -- that is slower than I would have expected.
Could you look up a couple of the metapaths that took that long? I'd like to investigate...

zietzm:
Thanks for the tips @kkloste ! I will need to be more comprehensive in order to see what the maximum dimension matrix in metapath is.

As far as simply runtime vs length, see below the figure with the average times by metapath length. Note that I define length as the number of edges. If I should use a different definition in the future for consistency, please let me know.



As a boxplot, we can see that only length 4 had the ~120 second runtimes. It looks like all the length 2 runtimes were very short (<2s), and that all length 3 were in the range ~13 seconds.

kkloste:
Nice work, @zietzm . It is comforting to know that only the length 4 paths are taking really long,
but the 120s time still concerns me. I suspect the long times occur because we're multiplying matrices in a sub-optimal order --- this is why I am curious to see the dimensions of the matrices underlying the metapaths that take 120s.

Example of what I am looking for:
if you multiplying matrices A, B, C, D all of dimensions 10x10, then the output is 10x10 and it takes O(10^3) work.
If instead their dimensions are (10x10^6) for A and C and (10^6x10) for B and D, then your output is still size 10x10, but the amount of work required varies drastically depending on how you perform your multiplies:

Doing (A*B) and (C*D) then (A*B)*(C*D) costs **O(10^8)** (the cost of multiplying together a 10x10^6 matrix and a 10^6x10 matrix).
However, doing (B*C) and then A*(B*C)*D costs **O(10^13)** because multiplying (B*C) makes a huge 10^6x10^6 matrix that you then have to deal with before things collapse down to the size 10x10.

My concern is that we might be performing this sub-optimal multiplying for some of the metapaths.
By looking at 5-10 metapaths of length 4 for which the time is 120s, I'll be able to confirm this theory, and then recommend a way forward.

zietzm:
Very interesting @kkloste. Here is a list of the 15 longest run-time metapaths. 

| metapath	 | length | time |
| --- | --- | --- |
| CuGiGcGuD	 | 4 | 133.82 |
| CuGcGiGuD	 | 4 | 132.53 |
| CbGiG<rGaD | 4 | 128.44 |
| CdGcGcGuD	 | 4 | 127.26 |
| CbGcGcGdD	 | 4 | 127.12 |
| CbGcGcGaD	 | 4 | 127.0 |
| CbGcGiGuD	 | 4 | 126.61 |
| CuGiGiGaD	 | 4 | 126.49 |
| CbGiGcGdD	 | 4 | 126.3 |
| CbGcGcGuD	 | 4 | 126.15 |
| CuGcGcGdD	 | 4 | 126.08 |
| CbGr>G<rGdD |	4 | 126.06 |
| CbGiGiGuD	| 4 | 125.99 |
| CbGr>G<rGaD | 4 | 125.99 |
| CbGiGiGaD | 4 | 125.94 |

What other kinds of information about these metapaths could I provide that would be helpful? @dhimmel suggested running a profiler to see where the computation time is spent. We discussed looking at how many `Gene ---> Gene` s there are in the metapath, as these are the largest matrices and could also be using a lot of time. The dimension of the largest matrix involved in the metapath is determined by the metanodes in question, so I will have a look.

kkloste:
All 15 of those have G-G-G in the middle part of the metapath --- that means that each of them involve a multiplication of a n-by-n matrix (for the first G-G edge) with another n-by-n matrix (for the second G-G edge), where n is roughly 20,000. That's causing a O(20,000^3) bottleneck!

As discussed in #53 , we can use matrix-chain-ordering to perform these operations in a more efficient way. The problem is that matrix-chain-ordering will do **all** the multiplies for a metapath, e.g. CuGiGcGuD, without giving us a chance to subtract the "loops matrix" in the interim.

To address this, I can return to the drawing board and try to concoct a different method for subtracting the "loops matrix" after the whole multiplication of the full path has been carried out.
This would mean using a two phase algorithm (1. compute full DWWC, 2. then subtract loops to get DWPC) instead of the iterative approach we currently implement (Do one multiply, subtract loops if needed, repeat).

zietzm:
@dhimmel The latest commits to this repository are a new run-through of the 1170 of 1206 Rephetio metapaths that do not require the general case dwpc function. As shown in the last cell of `CompareRephetio.ipynb`, our new method achieves identical results (to within 6 decimal places of the original results). In other words, we are now up to 1170 metapaths covered by specialized functions.




As you can see above, only a small number even took over 40 seconds this time. In the graph below, you can see that all of these paths have the segment `GeAeG` inside them.




As this graph shows, the small number of longest metapaths were the `GeAeG`s. The long bunching in `disjoint` is those paths involving `G-G-X-X`, generally.

zietzm:
I just finished the latest run-through of our metapaths. As it turns out, the updates we made today _somehow_ **cut the time in half**! I found this really surprising, but the paths all ran (and we got the same results) in about 45 minutes.

Anyway, I have not added the files to Git LFS because I am really not familiar with the software, and I had to head out. If we need these originals, they are on my computer here, and they can be regenerated quite quickly (now!).

I should be able to review some smallish changes in the next few days, but once the semester really picks up I don't imagine I'll have as much time.

Regardless, I really enjoyed working on hetmech this summer!

dhimmel:
> I just finished the latest run-through of our metapaths. As it turns out, the updates we made today somehow cut the time in half!

Nice!

I'm going to merge this and then make a PR that makes some additional changes.

PR #51, opened 2017-06-30
Multi duplicate

kkloste:
This Pull Request contains an outline of code for handling the BABA duplicate-nodes case.

There are two main components that need to be implemented:

1. a more sophisticated version of the current get_segments()
2. a straight-forward set of matrix operations to compute the DWPC once the correct segments are obtained.

(2) Will be simple. (1) won't be too difficult, but will require a bit more thought.


kkloste:
Handled by #52

PR #52, opened 2017-06-30
DWPC for metapaths with multi-repeated nodes

kkloste:
This Pull Request gives an example of why our previous methodology for DWPC with a single repeated node type will not work for the BABA case.
It contains an outline of code for handling the BABA duplicate-nodes case.

There are two main components that need to be implemented:

1. a more sophisticated version of the current get_segments()
2. a straight-forward set of matrix operations to compute the DWPC once the correct segments are obtained.

(2) Will be simple. (1) won't be too difficult, but will require a bit more thought.

### DIFF
--- a/hetmech/degree_weight_repeated_nodes_pseudocode.py
+++ b/hetmech/degree_weight_repeated_nodes_pseudocode.py
@@ -1,0 +1,75 @@
+def get_segments_multi(metagraph, metapath):
+    """
+    The intention is to call this method from "get_segments()"
+    in the case that overlapping segments with duplicate nodes are detected.
+
+    This function will then check for two possibilities:
+        (1) ABAB or BABA --- our code will handle these cases
+        (2) Anything else, i.e. BACBAC --- such cases raise an error
+
+    For ABAB
+        In the most general version of this, it is possible that there are
+        additional segments around and between the repeated ones, i.e.
+        UAXBYAZBV. Note that U, X, Y, Z, V can each represent entire segments
+        consisting of nodes that are not repeated anywhere else. A given
+        segment can include a single nodetype repeated, as long as no node in
+        any of these segments gets repeated in some other segment. So the
+        segment U could be nodes "DFGD" as long as D,F,G are not repeated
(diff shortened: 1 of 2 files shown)


kkloste:
@dhimmel @zietzm While thinking about implementing this, it occurred to me we ought to have a single `get_segments` function that determines what category the input metapath falls into:
1. No node repeats
2. Any number of nodes repeated once, but disjoint: `AXABYBCZC`
3. two nodes repeated once, nested: `BAAB`
4. two nodes repeated once, interleaved: `BABA`
5. any other situation

then `get_segments` returns the appropriate segments as well as a flag indicating which category the metapath is in. Then `dwpc()` calls the appropriate dwpc-subroutine to handle each segment, i.e. `dwpc_baab()`, and so on.

I'll work on a more generalized `get_segments()` function soon here. let me know if you have any input?

zietzm:
@kkloste I have implemented [(NB here)](https://github.com/zietzm/hetmech/blob/e3756dc6b0645f6dcde9f5a1c9f8e0fbe43f73b2/characterize_metapath.ipynb) [(explanation)](https://github.com/greenelab/hetmech/pull/59#issuecomment-316398881) some functionality  that may be a step towards your most recent comment. 

The functionality at this point is just to classify a metapath according to one of the categories we have set. I am not that familiar with the method and logic behind `get_segments`, though, so I am not sure I should work towards that bit until I better understand. 

Perhaps I should note too that I was working for a while on a faster way to do the exact DWPC calculations that I did in [this commit](https://github.com/zietzm/hetmech/blob/e3756dc6b0645f6dcde9f5a1c9f8e0fbe43f73b2/hetmech/dwpc_within_duplicated_metanode.py) but I haven't been able to find a quicker exact way. Even using sparse seems to take longer because we are dealing with vectors in the method. For some reason, sparse vector operations do a lot of dimensional stuff differently, leading me to have to implement much more time-intensive code. 

Anyway, here is a graph breakdown of the Rephetio metapaths using my most recent classification function:

---
metapath_categories

Issue #53, opened 2017-06-30
Optimize matrix chain multiplication


dhimmel:
@kkloste brought up that the order chained matrix multiplications are executed affects efficiency: https://github.com/greenelab/hetmech/pull/49#issuecomment-312283251. Turns out there's a [Wikipedia for this topic](https://en.wikipedia.org/wiki/Matrix_chain_multiplication).

Numpy has an implementation in [`numpy.linalg.multi_dot`](http://numpy.readthedocs.io/en/latest/reference/generated/numpy.linalg.multi_dot.html). If we dig through the [source code](https://github.com/numpy/numpy/blob/v1.11.0/numpy/linalg/linalg.py#L2362-L2400), there's a function `_multi_dot_matrix_chain_order`, which we could use to compute the optimal ordering.

We have to contend with the presence of mixed numpy and scipy.sparse arrays. Also, we may want to generate order just from the shape of the matrices, so we don't need all matrices in memory at the same time.

kkloste:
I agree, generate the order using only the shapes of the matrices first.

zietzm:
I have been running some profiles of various functions within degree_weight, and I have noticed that for dense matrices, the biggest time sink in many functions (eg: `degree_weight.dwpc`), is `functools.reduce`. This makes sense given the fact that some inefficient matrix multiplication ordering could be really slowing down our functions. I presume [this will be the line](https://github.com/greenelab/hetmech/blob/master/hetmech/degree_weight.py#L186) (among others) that we want to address with this issue.

However, it doesn't appear that matrix multiplication is a significant time sink when it comes to sparse matrices. Instead, several odd sparse-only operations appear to be the biggest time uses. For example, I just traced back through several functions and found that the biggest time use in `degree_weight.dwpc` with `sparse_threshold=1` (ie. sparse only throughout) was in conversion from `csr` to `csc` matrices. This seemed odd because we haven't been using `csr` at all (intentionally, at least). Turns out that in our current `dwwc_step`, performing degree-weighting on a `sparse.csc` adjacency matrix returns a `sparse.coo_matrix`. Subsequent multiplication of a `csc` and `coo` returns a `csr` matrix. 

If I add to the line in [matrix.auto_convert](https://github.com/greenelab/hetmech/blob/master/hetmech/matrix.py#L105) so that it reads
```python
    elif not sparse.issparse(matrix) and not above_thresh:
        return sparse.csc_matrix(matrix)
```
[...]

PR #54, opened 2017-07-05
Regularize sparse matrix type

zietzm:
Fixes the error whereby a `coo_matrix` (and then a `csr_matrix`) appears. This occurs when a `csc_matrix` is multiplied by a dense vector in `matrix.normalize`. 

CSC * Dense_vector = COO_matrix. 
COO * CSC = CSR

Conversion between sparse types is computationally expensive, so we want to ensure that we are working with a uniform sparse matrix type.

The premise of this method is the following:
Given:
`CSC_matrix * sparse column vector = CSR_matrix`

`CSC_matrix.transpose() = CSR_matrix`
`CSR_matrix * sparse row vector = CSR_matrix`
`CSR_matrix.transpose() = CSC_matrix`

Since in the numpy/scipy point-by-point multiplication scheme:
matrix * column vector = (matrix.T * row vector).T

Then we use
`(CSC.T * row_vector_of_row_sums).T`
as a substitute for 
`CSC(CSC * column_vector_of_row_sums)`, which would involve a costly type conversion.

Computing the transpose is not free, but this method works a good bit faster for me than doing a conversion between matrix types.

Referenced in #53

### DIFF
--- a/hetmech/diffusion.py
+++ b/hetmech/diffusion.py
@@ -38,7 +38,7 @@ def diffusion_step(
     numpy.ndarray
         Normalized matrix with dtype.float64.
     """
-    # returns a newly allocated array
+    # returns a newly allocated array if copy is True
     matrix = copy_array(matrix, copy=copy)
 
     # Perform column normalization
--- a/hetmech/diffusion.py
+++ b/hetmech/diffusion.py
@@ -9,13 +9,14 @@ from .matrix import (normalize,
 
 
 def diffusion_step(
-        matrix, row_damping=0, column_damping=0, copy=True):
+        matrix, row_damping=0, column_damping=0):
     """
     Return the diffusion adjacency matrix produced by the input matrix
     with the specified row and column normalization exponents.
     Note: the row normalization is performed second, so if a value
     of row_damping=1 is used, the output will be a row-stochastic
-    matrix regardless of choice of column normalization.
+    matrix regardless of choice of column normalization. Matrix will
+    not be modified in place.
 
     Parameters
     ==========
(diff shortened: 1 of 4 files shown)


zietzm:
As mentioned in #53, the fact that the `copy` argument does not work the same way for numpy arrays and scipy sparse matrices means that `test_diffusion` will return an error.

The difference between numpy array and sparse matrix is as follows:
```python
mat = numpy.array([
    [1, 2],
    [3, 4]
])
csc = sparse.csc_matrix(mat)

mat_copy = numpy.array(mat, copy=False)
mat_copy is mat
>>> True

csc_copy = sparse.csc_matrix(csc, copy=False)
csc_copy is csc
>>> False

type(csc_copy) is type(csc)
>>> True
```

While we now don't have to worry about the expensive step of converting between matrix types, this method is not perfect. 

In the important case where we have a sparse matrix, with `copy=False` and `dtype=numpy.float64`, therefore, we would now be lacking the ability to normalize the matrix in place.

zietzm:
@kkloste suggested that we use the vector in sparse diagonal form. This speeds things up even more than the original proposed method.

Matrix-vector multiplication is now as follows:
```python
vector = sparse.diags(vector)
if axis == 'rows':
    matrix = (matrix.T @ vector).T
else:
    matrix = matrix @ vector
```

hetmech/diffusion.py
```diff
@@ -9,13 +9,14 @@
 
 
 def diffusion_step(
```
dhimmel: This can now all be on one line?

PR #57, opened 2017-07-07
Dual normalize corner cases

### LINKED ISSUE #22: dual_normalize test should include NaN / Inf corner cases
A good test of dual_normalize should also check it handles properly the corner case that a column or row is empty, or that the normalization procedures somehow lead to NaN or Inf

zietzm:
Closes #22 

Add a a test in `test_diffusion` to test corner cases like zero matrix, matrix with empty row, empty matrix, matrix with NaN entries, matrix with Inf entries. Adds lines in `matrix.copy_array` to raise errors in the cases which should not be computed.


data/.gitignore
```diff
@@ -1 +1,3 @@
 xarray_dataset.pkl.gz
+dwpc_metapaths.tsv
```
dhimmel: These ignores don't belong with this PR?

hetmech/matrix.py
```diff
@@ -109,9 +109,17 @@ def auto_convert(matrix, threshold):
...
 def copy_array(matrix, copy=True):
     """Returns a newly allocated array if copy is True"""
+    assert matrix.ndim == 2
+    assert matrix.dtype != 'O'  # Ensures no empty row
+    if not sparse.issparse(matrix):
+        assert numpy.isfinite(matrix).all()  # Checks NaN and Inf
+    try:
+        matrix[0, 0]  # Checks that there is a value in the matrix
+    except IndexError:
+        raise AssertionError
```
dhimmel:
```
raise AssertionError('Human readable error message here)
```

Issue #58, opened 2017-07-10
Coordinating matrix DWPC implementation with Scripps folks


dhimmel:
Several individuals in the [Su Lab](http://sulab.org/) at Scripps Research Institute, specifically @veleritas, @NuriaQueralt and @mmayers12, are working on extensions to Rephetio and Hetionet. We're going to catch up on a call Wednesday July 12, 2017 from 1-3 PM. As an aside, @cgreene @zietzm @danich1 @kkloste, feel free to join this call, even if just as background audio. Since we want to focus the call on more high level planning, @mmayers12 suggested we take a look at his work on implementing matrix DWPC beforehand.

You can see the current progress in the [`mmayers12/hetnet-ml`](https://github.com/mmayers12/hetnet-ml) repo and specifically in [`matrix_tools.py`](https://github.com/mmayers12/hetnet-ml/blob/c330f1bd81ca33fb7f0314bed16c5af6f4e72930/src/matrix_tools.py). It looks like @mmayers12 has arrived to several of the same decisions and algorithms we've implemented here. For example, the DWWC and DWPC distinction.

For reference, the work in hetmech currently consists of three primary contributors:

+ @kkloste, an algebraic mathematician from North Carolina State University. Kyle helps us with figuring out how to do things using matrices.
+ @zietzm, who's interning in the Greene Lab this summer, does much of the Python implementation. He's now the most knowledgeable about matrices in Python, and has a good handle on the `scipy.sparse` API.
+ @dhimmel (me), who's main involvement is with the application of these methods to hetnets and strategic directions.

It would be great if @mmayers12, @kkloste, @zietzm and I could all get on the same page regarding how the `mmayers12/hetnet-ml` and `greenelab/hetmech` implementation differ. My hope is that @mmayers12 will be able to contribute his advances to this repo. Specifically, we should compare our matrix-DWPC algorithms which are discussed specifically in #52, #47, #45 (merged), and #20 (outdated). Our current implementation is in [`degree_weight.py`](https://github.com/greenelab/hetmech/blob/a2cc88ce2e28a5108d528f95470347c32a6d5c85/hetmech/degree_weight.py#L163-L187). In short, @kkloste has figured out how to exclude duplicate-node paths for some metapaths but not all.

[...]

PR #59, opened 2017-07-13
[WIP] Add general DWPC method for arbitrary metapaths

zietzm:
So after the issues we encountered yesterday, I did some thinking about ways to get the exact path solutions when we have repeated metanodes. To restate the issue we had been having:

We have eliminated many of the issues with cases of repeated metanode A-B-C-B, etc. when dealing with metanodes (capital letters here). The problem we had was with repeated metanodes, A-A-A-A that we would be counting the paths of nodes (lowercase letters here) a-b-c-b, which shouldn't have been counted as they weren't paths. 

I mentioned this in [this comment](https://github.com/greenelab/hetmech/issues/47#issuecomment-314571953) and @kkloste discussed it with some of the Scripps folks in #58. 

This PR is a potential implementation of path counting that should eliminate this issue. I think it should only be necessary in the aforementioned A-A-[...]-A metapath with repeated metanodes, as I'm sure this will be considerably slower than the standard matrix implementation (I plan on verifying this sometime this afternoon). It sounds like this could could potentially work together with the generalized `get_segements` function you mentioned [here](https://github.com/greenelab/hetmech/pull/52#issuecomment-315139086)

The basic idea of this method is in the image below: 

scheme

1. Select a start node and a history (in the first node this is just a ones vector). 
2. Subtract the start vector from the history vector
3. Multiply the start vector by the adjacency matrix
4. Element-wise multiply the history vector by the result of (3.)
5. By creating a diagonal matrix from the resulting vector and selecting the rows,  get a set of vectors that each have just one element-- child nodes.
6. Repeat (1-5) with each resulting vector and the current history vector until the desired depth is reached.
7. Sum the resulting vectors to get path targets for the queried vector.

An example is the simple graph below:

example

My thinking on this is the following: this will be a slow method. We have to create a new vector for every single fork in the path. Depending on how this performs on an example case, I may determine that this is not a computationally feasible approach as it takes far too much time and/or memory. However, on the off-chance that this is a feasible method, it would be nice to be getting exact results for path count under conditions of repeated metanode. 

Even if this isn't feasible, perhaps the work in this PR could in some way be helpful, at least for thinking about approaches to the A-A-[...]-A repeated metanode in the future.


zietzm:
I should note, too that as shown, and as in the code, this is purely a path count, not a DWPC. Though, if this works, that should not be a difficult addition at all.

zietzm:
I just ran the `CrCrCrCtD` metapath with the new method, and I can confirm that it is considerably slower than the previous method (it took almost 30 minutes). 

The last comparison I did was between the Rephetio results and the previous method. The Rephetio data was a good bit smaller than the hetmech data at the time, and we had disagreement about 447 and agreement about 3328 data points. With the new method, **we correctly got all 3375 data points!**

So, perhaps it is now possible that with the combination of an improved `get_segments` function that does some classification of the metapath and metapath matrix caching, we will be able to calculate the DWPC for any two query nodes **very** quickly. 

I imagine it will take some time to come up with a better `get_segments` function which can classify the metapaths over which the new method should be used (and the results cached). In the meantime, I hope to improve my new functionality so that it uses sparse matrices and sparse vectors. This should make the calculation considerably faster.

zietzm:
The last five commits to this PR added [this notebook](https://github.com/zietzm/hetmech/blob/e3756dc6b0645f6dcde9f5a1c9f8e0fbe43f73b2/characterize_metapath.ipynb), which contains three relevant functions for characterizing metapaths. 


1. The first removes all extraneous nodes and all edges.

'AxByBxCzA' -> ['A', 'B', 'B', 'A']


2. The second outputs the smallest segments which can be individually characterized

'AxByBxCzA' -> ['ABBA']
'AxAyBzBaCbC' -> ['AA', 'BB', 'CC']
'AxByCzD' -> ['']


3. The third characterizes output from the second.

Returns one of the following: ['no_repeats', 'disjoint', 'abba', 'abab', 'Other']

I still need to trim these functions down but I think the method is there.

hetmech/test_matrix.py
```diff
@@ -71,12 +71,33 @@ def test_metaedge_to_adjacency_matrix(test_edge, dtype, threshold):
...
+                                      ('TlDaGeTlDaG', 0),  # ABCABC
+                                      ('DaGiGiGiGiGaD', None),  # ABBBBBA
+                                      ('GiGiGaDlTlD', 'disjoint'),
+                                      ('GiGiGiGiGiGiGiGiG', 'disjoint'),
+                                      ('DlTlDlTlD', 'other')])  # ABABA
 def test_categorize(metapath):
-    assert categorize(metapath[0]) == metapath[1]
+    url = 'https://github.com/dhimmel/hetio/raw/{}/{}'.format(
+        '9dc747b8fc4e23ef3437829ffde4d047f2e1bdde',
+        'test/data/disease-gene-example-graph.json',
```
dhimmel: Just read in the metagraph `test/data/hetionet-v1.0-metagraph.json` and not the subgraph

hetmech/matrix.py
```diff
@@ -123,3 +124,55 @@ def copy_array(matrix, copy=True):
...
+    # Group neighbors if they are the same
+    grouped = [list(v) for k, v in itertools.groupby(repeats_only)]
+
+    # Handle multiple disjoint repeats, any number, ie. AA,BB,CC,DD,...
+    if len(grouped) == len(repeated_nodes):
+        return 'disjoint'
+
+    # Group [A, BB, A] or [A, B, A, B] into one
+    if len(repeats_only) - len(grouped) <= 1:
+        grouped = [[item for sublist in grouped for item in sublist]]
```
dhimmel: is this the same as `[list(itertools.chain(*grouped))]`?

Is this also the same as `[repeats_only]`?

hetmech/matrix.py
```diff
@@ -123,3 +124,55 @@ def copy_array(matrix, copy=True):
...
+            return 'BAAB'
+        else:
+            return 'BABA'
+    else:
+        # Multi-repeats that aren't disjoint, eg. ABCBAC
+        if len(repeated_nodes) > 2:
+            raise NotImplementedError(
+                "Only two overlapping repeats currently supported")
+
+        if len(metanodes) > 5:
```
dhimmel: Are we sure there won't be errors withholding this check so late?

hetmech/matrix.py
```diff
@@ -123,3 +124,55 @@ def copy_array(matrix, copy=True):
...
+            raise NotImplementedError(
+                "Metapaths of that length are not yet supported")
```
dhimmel: Be more specific:

"Metapaths of length > 4 are not yet supported"

hetmech/matrix.py
zietzm: Yeah, it is the same as both, and I think it definitely makes more sense to just use `[repeats_only]`

hetmech/matrix.py
zietzm: The only reason I do is because we don't have to add any more functionality to correctly categorize `AABBCCDD...`, an arbitrary number of disjoint repeats. Most of the code above that is just to adequately cover the disjoint case. It could be move up somewhat, though, if you think it would be useful.

zietzm:
Just a quick note:

---
As of #62, this is the most complex DWPC scheme of the ones to go into dwpc.py. Nonetheless, this does not include any support for random metanode inserts into long_repeats. For example `A-A-B-A-C-A` is not supported by this scheme. 

There exists a most general scheme for doing this, but it is even slower than the method in this PR. It may make sense to do that as a final dwpc pull request, as the method would be generalize-able to every metapath, regardless the length or composition. It would make sense to put this into a final PR to dwpc.py as it would take a good while to develop and could be used for the `else` case at the end of the dwpc function to catch all metapaths that were not able to be processed by any of the other methods. There would also then be easy work for someone to write individual functions to address the specific metapath types which take the longest.

zietzm:
Closed by #67 in branch develop

PR #60, opened 2017-07-21
Add metapath categorization

zietzm:
Adds a function `categorize` and tests for it. Takes a `hetio.hetnet.MetaPath` object and returns a string of one of the following:`['no_repeats', 'disjoint', 'BAAB', 'BABA', 'other']`

A question I have is how we should handle `GiGcGcGiG`. Currently, I have this being categorized as `other`, not as `BAAB`, but it could almost go either way. There are two adjacency matrices involved, so I am tempted to say this should be a `BAAB`.


kkloste:
@zietzm and @dhimmel -- I apologize for being out of the loop for so long; I was out of town for a conference last week, and preparing for it the week before.

One comment: `GiGcGcGiG` should be classified as "other" because even though the particular adjacency matrices change, the node-set (`G`) is the same, and so this metapath could conceivably contain a path with as many as 4 loops, unless I misunderstand the labeling notation @dhimmel ?

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+    GiGiG   -> 'short_repeat'
+    GiGiGcG -> 'long_repeat'
+    GiGcGiG -> 'long_repeat'
+    GiGbCrC -> 'disjoint'
+    GbCbGbC -> 'BABA'
+    GbCrCbG -> 'BAAB'
+    DaGiGbCrC -> 'disjoint'
+    GiGaDpCrC -> 'disjoint'
+    GiGbCrCpDrD -> 'disjoint'
+    GbCpDaGbCpD -> NotImplementedError
```
dhimmel: update to `other`

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+    GbCrCrCrCrCbG -> NotImplementedError
+    """
+    metanodes = list(metapath.get_nodes())
+    repeated_nodes = {v for i, v in enumerate(metanodes) if
```
dhimmel: Let's change this to `repeated_metanodes` or `repeated`

We want to help readers disambiguate node and metanode

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+
+    repeats_only = [node for node in metanodes if node in repeated_nodes]
+
+    # Group neighbors if they are the same
+    grouped = [list(v) for k, v in itertools.groupby(repeats_only)]
+
+    # Handle multiple disjoint repeats, any number, ie. AA,BB,CC,DD,...
+    if len(grouped) == len(repeated_nodes):
+        # Identify if there is only one metanode
+        if len(set(repeated_nodes)) == 1:
```
dhimmel: repeated_nodes is already a set

hetmech/dwpc.py
dhimmel: Now that you have the counter, it may be more straightforward to just do:

```python
{metanode for metanode, count in freq.items() if count > 1}
```

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+            if max(freq.values()) < 4:
+                return 'short_repeat'
+            else:
+                return 'long_repeat'
+
+        return 'disjoint'
+
+    # Categorize the reformatted metapath
+    if len(repeats_only) in (3, 4):
```
dhimmel: You can do this if you want

```python
if 3 <= len(repeats_only) <= 4:
```

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+    if len(grouped) == len(repeated_nodes):
+        # Identify if there is only one metanode
+        if len(set(repeated_nodes)) == 1:
+            if max(freq.values()) < 4:
+                return 'short_repeat'
+            else:
+                return 'long_repeat'
+
+        return 'disjoint'
+
```
dhimmel:
```python
assert len(repeats_only) > 3
```

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+    # Categorize the reformatted metapath
+    if len(repeats_only) in (3, 4):
+        if repeats_only[0] == repeats_only[-1]:
+            return 'BAAB'
+        else:
+            return 'BABA'
```
dhimmel: Let's add the assertion for doc/safety:

```python
assert repeats_only[0] == repeats_only[2] and repeats_only[1] == repeats_only[3]
```

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+    elif len(repeats_only) == 5 and max([len(i) for i in grouped]) == 3:
```
dhimmel: A bit shorter: `max(map(len, grouped))`

hetmech/dwpc.py
```diff
@@ -204,15 +207,176 @@ def dwpc_long_repeat(graph, metapath, damping=0.5):
...
+            logging.info(
+                f"{metapath}: Only two overlapping repeats currently supported"
+            )
+
+        if len(metanodes) > 5:
+            logging.info(
+                f"{metapath}: Complex metapaths of length > 4 are not yet "
+                f"supported")
+        return 'other'
+
```
dhimmel: Should you `assert False` at end of function since you are assuming you will never reach this point?

zietzm: That would work if I did `return 'other'` above in the two logged cases. Otherwise the metapaths that will get logged there will not be returned.

zietzm: But come to think of it, that may actually make the code in this region both more readable and easier to change in the future if either of the logged conditions is changed by a new DWPC function.

zietzm:
Moved to develop branch #65

PR #61, opened 2017-07-24
Add DWPC functionality for BABA case

zietzm:
This is a similar procedure to #59, in that it uses a history. However, unlike that method, this should be faster, as it stores the history as only a single element. This leads to this PR being functional only for the BABA case.


zietzm:
In reference to the `BAAB` case, @kkloste, it seems that this is fairly trivial, requiring nothing more than subtracting the diagonal after every repeated segment. Am I missing something here (in the `BAAB` case)?

For example, say we have three metanodes, A, B, and C, with the corresponding adjacency matrices ba, ac, cc. The following seems to me to give the correct path count on an example graph I made. 
```python
def remove_diag(mat):
    return mat - np.diag(mat.diagonal())

remove_diag(ba@aa@ba.T)  # BAAB
remove_diag(ba@remove_diag(ac@cc@ac.T)@ba.T)  # BACCAB
```
I get the sense that there may be something more complicated going on that I have simply missed. I know that the overlapping repeats (as in `BABA`) are not so simply removed.

kkloste:
@zietzm -- to answer your question from 8 days ago:

> In reference to the `BAAB` case, @kkloste, it seems that this is fairly trivial, requiring nothing more than subtracting the diagonal after every repeated segment. Am I missing something here (in the `BAAB` case)?

For the `BXAYAZB` case we would:

* compute `AYA`, then subtract its diagonal to get `M_1`
* then compute `BX M_1 ZB`, and subtract its diagonal.

In these examples, `X`, `Y`, and `Z` can each represent metapaths of any length that do not contain any repeats (and do not contain metanodes `A` or `B`.

```python
remove_diag(ba@aa@ba.T)  # BAAB
remove_diag(ba@remove_diag(ac@cc@ac.T)@ba.T)  # BACCAB
```
I think this is correct, it depends on your notation --- are the innermost objects `aa` and `cc` single matrices, or possibly multiple matrices multiplied together?

kkloste:
Even if `cc` represents a "single matrix" i.e. a set of edges going from metanode `c` to metatype `c`,
if there exist any loop edges in the matrix then it must still have its diagonal removed.

kkloste:
Hopefully @zietzm this pseudo code makes clear what I mean:

img_20170801_200041-1

zietzm:
Moved to develop branch #65

PR #62, opened 2017-07-26
New file for DWPC

zietzm:
This is a new layout file specifically for all the DWPC functions, so that they can be organized nicely, and so that it is easy to see where work has been done and still needs to be done. I would plan to rebase my other open pull requests,  #59, #60, and #61 such that their functions are being put here. I think it definitely makes less sense to have each of those pull requests creating a new file when we will want to bring them all together anyway.

### DIFF
--- a/hetmech/dwpc.py
+++ b/hetmech/dwpc.py
@@ -8,7 +8,7 @@ from .matrix import metaedge_to_adjacency_matrix, normalize, copy_array
 
 def remove_diag(mat):
     """Set the main diagonal of a square matrix to zeros."""
-    assert len(set(mat.shape)) == 1  # must be square
+    assert mat.shape[0] == mat.shape[1]  # must be square
     return mat - numpy.diag(mat.diagonal())
 
 
@@ -27,14 +27,14 @@ def dwpc_no_repeats(graph, metapath, damping=0.5):
     assert len(set(metapath.edges)) == len(metapath)
 
     parts = list()
-    for edge in metapath:
+    for metaedge in metapath:
         rows, cols, adj = metaedge_to_adjacency_matrix(
-            graph, edge, dtype=numpy.float64, sparse_threshold=0)
+            graph, metaedge, dtype=numpy.float64, sparse_threshold=0)
         adj = degree_weight(adj, damping)
 
-        if edge == metapath[0]:
+        if metaedge == metapath[0]:
             row_names = rows
-        if edge == metapath[-1]:
+        if metaedge == metapath[-1]:
             col_names = cols
         parts.append(adj)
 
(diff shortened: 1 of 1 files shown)


hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+
+import numpy
+
+from .matrix import metaedge_to_adjacency_matrix, normalize, copy_array
+
+
+def remove_diag(mat):
+    """Set the main diagonal of a square matrix to zeros."""
+    assert len(set(mat.shape)) == 1  # must be square
+    return mat - numpy.diag(mat.diagonal())
```
dhimmel: I thought we had decided to always subtract a sparse diagonal for performance / minimizing RAM? Were there type issues? Just a note, can deal with this later.

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+    matrix = normalize(matrix, column_sums, 'columns', damping)
+
+    return matrix
+
+
+def dwpc_no_repeats(graph, metapath, damping=0.5):
+    assert len(set(metapath.edges)) == len(metapath)
+
+    parts = list()
+    for edge in metapath:
```
dhimmel: for clarity, let's go for `metaedge in metapath`.

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def dwpc_short_repeat(graph, metapath, damping=0.5):
+    """
+    One metanode repeated 3 or fewer times (A-A-A), not (A-A-A-A)
+    This can include other random inserts, so long as they are not
+    repeats. Must start with the repeated node. Acceptable examples:
+    A-B-A-A) (A-B-A-C-D-E-F-A) (A-B-A-A-C), etc.
+    """
+    start_node = metapath.get_nodes()[0]
```
dhimmel: `start_metanode` not `start_node`. Can also [use](https://github.com/dhimmel/hetio/blob/30c6dbb18a17c05d71cb909cf57af7372e4d4908/hetio/hetnet.py#L142-L143) `metapath.source()`.

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+    dwpc_matrix = None
+    for edge in metapath:
+        rows, cols, adj = metaedge_to_adjacency_matrix(
+            graph, edge, dtype=numpy.float64, sparse_threshold=0)
+        adj = degree_weight(adj, damping)
+        if dwpc_matrix is None:
+            row_names = rows
+            dwpc_matrix = adj
+        else:
+            dwpc_matrix = dwpc_matrix@adj
```
dhimmel: `dwpc_matrix @ adj` spaces

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+
+    parts = list()
+    for edge in metapath:
+        rows, cols, adj = metaedge_to_adjacency_matrix(
+            graph, edge, dtype=numpy.float64, sparse_threshold=0)
+        adj = degree_weight(adj, damping)
+
+        if edge == metapath[0]:
+            row_names = rows
+        if edge == metapath[-1]:
```
dhimmel: This is nice and explicit, but you can always just returns cols, which will be set from the last iteration of the loop. Your choice, since sometimes being clear is more important than being brief.

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+    for edge in metapath:
+        rows, cols, adj = metaedge_to_adjacency_matrix(
+            graph, edge, dtype=numpy.float64, sparse_threshold=0)
+        adj = degree_weight(adj, damping)
+        if dwpc_matrix is None:
+            row_names = rows
+            dwpc_matrix = adj
+        else:
+            dwpc_matrix = dwpc_matrix@adj
+        if edge.target == start_node:
```
dhimmel: So now we're removing after every step?

hetmech/dwpc.py
```diff
@@ -0,0 +1,108 @@
...
+import operator
+
+import numpy
+
+from .matrix import metaedge_to_adjacency_matrix, normalize, copy_array
+
+
+def remove_diag(mat):
+    """Set the main diagonal of a square matrix to zeros."""
+    assert len(set(mat.shape)) == 1  # must be square
```
dhimmel: I prefer `mat.shape[0] == mat.shape[1]`

hetmech/dwpc.py
zietzm: Just before you reviewed this, it came to my attention that this function works for ABAA but not for AABA. I will have to do some segmentation for it to work better.

hetmech/dwpc.py
```diff
@@ -0,0 +1,127 @@
...
+
+import numpy
+
+from .matrix import metaedge_to_adjacency_matrix, normalize, copy_array
+
+
+def remove_diag(mat):
+    """Set the main diagonal of a square matrix to zeros."""
+    assert mat.shape[0] == mat.shape[1]  # must be square
+    return mat - numpy.diag(mat.diagonal())
```
kkloste: When the underlying matrix is large this is inefficient; the diagonal matrix should be sparse.

PR #70, opened 2017-08-23
Merge 'Develop' into Master

zietzm:
This PR includes all the DWPC functions necessary to completely replicate the results in Rephetio (see #49). Additions since the last changes to master include additional metapath categorizations which allow several longer metapaths to be classified accurately. 

Note that metapaths of length six (six edges, seven metanodes) are only partially supported by the current work. While in principle all metapaths are supported with the general function, there are not yet any sort of specialized functions analogous to `dwpc_baba` for length six metapaths. That is not to say, however, that no 6-paths are supported. In fact, many which are simply disjoint repeats and repeats around pre-defined groups are entirely supported. Additionally, since length six cannot introduce an extra *pair* of repeats from length five, the only added difficulty for these paths should lie in potentially longer split repeats which would need a form of merged `BABA` and `short_repeat` DWPC functions.

In addition to the general case DWPC function (#59 #67), this PR incorporates the work @kkloste did in #52 towards the `BABA` case, as well as functions for `short_repeat`, `disjoint`, and `BAAB`. 

I have included here additional work towards `categorize` and `get_segments` which have been significantly expanded. These functions should hopefully be organized in a logical way which will allow relatively simple expansion to further metapaths (especially those of length 6). 

Finally, a quick note about the test file:
There are no explicit tests for the `disjoint` case due to the fact that the example graph we use in test files does not have any `DrD`, `CrC`, etc. metaedges. The only repeat in the example is `GiG`. My reason for not including a test for `disjoint` is that the example graph does not have the capability to have disjoint metapaths.

Closes https://github.com/greenelab/hetmech/pull/52
Closes https://github.com/greenelab/hetmech/issues/20


zietzm:
Just to note, I have moved all DWPC-associated functions to the degree_weight.py and test_degree_weight.py files. Besides a few updates I made to test_path_count.py, in order to keep them up to date with the latest DWPC function outputs, all relevant changes should by localized to degree_weight and its corresponding test function.

kkloste:
Great work, @zietzm !
If I understand this correctly, then every metapath on 5 or fewer edges can now be handled?

The only other idea I have for improving things --- @zietzm the method you devised for handling one of the more complex metapath types, if I recall, you had said was a little slow? (I'm referring to the method involving "history vectors" implemented here https://github.com/greenelab/hetmech/pull/67/commits/cf21a818d4bb5f2f06fedc488bce7e9048eb0faa)

My understanding is this subroutine is not called often, so I don't know if it is worth additional effort to speed this up?
If it is worth the effort, I know how we can further generalize the previous work for handling repeated nodes so that we compute this using just simply matrix multiplies and subtracts.
@dhimmel let me know if it'll be worth the time.

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+            else:
+                dwpc_matrix = dwpc_matrix @ seg_matrix
+
+    elif category == 'repeat_around':
+        mid = dwpc(graph, segments[1], damping=damping)[2]
+        row, c, adj0 = _dwpc_no_repeats(graph, segments[0], damping=damping)
+        r, col, adj1 = _dwpc_no_repeats(graph, segments[-1], damping=damping)
+        dwpc_matrix = _remove_diag(adj0 @ mid @ adj1)
+
+    elif category == 'short_repeat' and len(segments) != 1:
```
dhimmel: `len(segments) > 1`

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+            row, col, dwpc_matrix = _dwpc_general_case(
+                graph, metapath, damping)
+        else:
+            raise Exception('Metapath category will use _dwpc_general_case')
+
+    else:
+        row, col, dwpc_matrix = category_to_function[category](
+            graph, metapath, damping, sparse_threshold=sparse_threshold)
+    total_time = time.perf_counter() - start_time
+    return row, col, dwpc_matrix, total_time
```
dhimmel: Consistent row / col nomenclature

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+            if segment is segments[-1]:
+                col = c
 
-    return matrix
+            if dwpc_matrix is None:
+                dwpc_matrix = seg_matrix
+            else:
+                dwpc_matrix = dwpc_matrix @ seg_matrix
+
+    elif category == 'repeat_around':
```
dhimmel: Consider making functions for these.

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+    matrix = copy_array(matrix, copy)
+    row_sums = numpy.array(matrix.sum(axis=1)).flatten()
+    column_sums = numpy.array(matrix.sum(axis=0)).flatten()
+    matrix = normalize(matrix, row_sums, 'rows', damping)
+    matrix = normalize(matrix, column_sums, 'columns', damping)
+
+    return matrix
+
+
+def _dwpc_no_repeats(graph, metapath, damping=0.5, sparse_threshold=0):
```
dhimmel: Remove and use dwwc instead

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
     """
-    # returns a newly allocated array if copy is True
-    matrix = copy_array(matrix, copy)
+    start_time = time.perf_counter()
+    category_to_function = {'no_repeats': _dwpc_no_repeats,
+                            'short_repeat': _dwpc_short_repeat,
+                            'long_repeat': _dwpc_general_case,
+                            'BAAB': _dwpc_baab,
+                            'BABA': _dwpc_baba,
+                            'complex': dwpc,
```
dhimmel: `complex` no longer returned by categorize.

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+        element.
+
+    Examples
+    --------
+    'CbGaDaGaD' -> ['CbG', 'GaD', 'GaG', 'GaD']
+    'GbCpDaGaD' -> ['GbCpD', 'DaG', 'GaD']
+    'CrCbGiGaDrD' -> ['CrC', 'CbG', 'GiG', 'GaD', 'DrD']
     """
+
+    def add_head_tail(metapath, indices):
```
dhimmel: Docstring. Consider move outside function fro readability.

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+    Parameters
+    ----------
+    metagraph : hetio.hetnet.MetaGraph
+    metapath : hetio.hetnet.Metapath
+
+    Returns
+    -------
+    list
+        list of metapaths. If the metapath is not segmentable or is already
+        fully simplified (eg. GaDaGaD), then the list will have only one
```
dhimmel: Outdated doc.

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+    return segments
+
+
+def _remove_diag(mat):
+    """Set the main diagonal of a square matrix to zeros."""
+    mattype = type(mat)
+    if mattype == numpy.ndarray:
+        mattype = numpy.array
+    assert mat.shape[0] == mat.shape[1]  # must be square
+    return mat - mattype(numpy.diag(mat.diagonal()), dtype=numpy.float64)
```
dhimmel: Consider [this method](https://github.com/greenelab/hetmech/blob/775ef9b38f549dfd147104c66a256ae9c7df4605/hetmech/degree_weight.py#L155-L157)

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+            seg_bed = segments[-1] if segments[-1] != seg_azb else None
+    # Collect segment DWPC and corrections
+    row_names, col, axb, seconds = dwpc(graph, seg_axb, damping=damping,
+                                        sparse_threshold=sparse_threshold)
+    row, col, bya, seconds = dwpc(graph, seg_bya, damping=damping,
+                                  sparse_threshold=sparse_threshold)
+    row, col_names, azb, seconds = dwpc(graph, seg_azb, damping=damping,
+                                        sparse_threshold=sparse_threshold)
+
+    correction_a = numpy.diag((axb @ bya).diagonal()) @ azb
```
dhimmel: Would a sparse solution be best?

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
+                                            sparse_threshold=sparse_threshold)
+        dwpc_matrix = cda @ dwpc_matrix
+    if seg_bed is not None:
+        row, col_names, bed, seconds = dwpc(graph, seg_bed, damping=damping,
+                                            sparse_threshold=sparse_threshold)
+        dwpc_matrix = dwpc_matrix @ bed
+    return row_names, col_names, dwpc_matrix
+
+
+def _dwpc_short_repeat(graph, metapath, damping=0.5, sparse_threshold=0):
```
dhimmel: Reminder

hetmech/degree_weight.py
```diff
@@ -1,187 +1,642 @@
...
-        Normalized matrix with dtype.float64.
+    -------
+    numpy.ndarray, numpy.ndarray, numpy.ndarray or
+    scipy.sparse.csc_matrix, float
+        row labels, column labels, the DWPC matrix, and the computation time
     """
-    # returns a newly allocated array if copy is True
-    matrix = copy_array(matrix, copy)
+    start_time = time.perf_counter()
+    category_to_function = {'no_repeats': _dwpc_no_repeats,
```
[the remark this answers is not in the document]
zietzm: Tried it, and it gives errors because the functions have to all be above the dictionary definition. Since it is small, I think it may be best to leave it inside here.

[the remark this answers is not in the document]
zietzm: Actually, since some of the previous sub-functions need it, I may just make it into its own function.

PR #72, opened 2017-08-29
Refactor Rephetio DWPC notebooks

dhimmel:
Follows on https://github.com/greenelab/hetmech/pull/49. Refactors notebooks @zietzm created during his summer Vagelos Scholars Program (reported on [here](https://zietzm.github.io/Vagelos2017/) & on [figshare](https://doi.org/10.6084/m9.figshare.5346577)).

@zietzm check out [`6.compare-rephetio.ipynb`](https://github.com/dhimmel/hetmech/blob/caf6c8187ce715343ae4c286d16a79a35e1b23d4/6.compare-rephetio.ipynb). Looks like we have a bug where our matrix DWPC implementation is overestimating the DWPC. It appears that the bug affects BBAB style metapaths.

Would be nice if we could fix the bug and then rebase this PR on the fixed DPWC code.

PR #73, opened 2017-08-31
Fix short_repeat error

zietzm:
Refs https://github.com/greenelab/hetmech/pull/72#issue-253793454


zietzm:
@dhimmel I think I have corrected the error in short_repeat, but I cannot get the Travis build to succeed. I am really not sure why this is, as `pytest hetmech` runs through all successfully on my machine.
Is there any chance you could take a look at the Travis errors to see what is going on?
It seems to me like this should all be working fine, as I think I fixed all the non-notebook flake8 errors.

hetmech/test_short_repeat_error.py
```diff
@@ -0,0 +1,36 @@
...
+    ('DB00571', 'DOID:1826', 'CtDrDaGaD', 0.008234),
+    ('DB00996', 'DOID:1826', 'CtDrDaGaD', 0.000395),
+    ('DB00313', 'DOID:1826', 'CtDrDaGaD', 0.000395)
+])
+def test_short_error(error_compound, error_disease, metapath, cypher):
+    commit = '59c448fd912555f84b9822b4f49b431b696aea15'
+
+    url = (f'https://github.com/dhimmel/hetionet/raw/{commit}/hetnet/json/'
+           f'hetionet-v1.0.json.bz2')
+    graph = hetio.readwrite.read_graph(url)
```
dhimmel: Rereading hetionet each time is slow. Even if you fixed this to read hetionet only once, that's still quite a heavy test.

Let's add a `--runslow` pytest option as per https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option and mark `test_short_error` as slow.

zietzm:
I tried to add it, but I now realize that it wasn't really necessary to begin with. As long as the new method gets the correct results we shouldn't really need an additional test file. It has been deleted.

PR #77, opened 2018-02-19
Add demonstration notebook for gene queries

dhimmel:
Search for biological processes implicated by blood pressure gene expression along the GiGpBP metapath.


dhimmel:
Rerunning with fix with:

```
jupyter nbconvert --inplace --execute \
  --ExecutePreprocessor.timeout=-1 \
  --ExecutePreprocessor.kernel_name=python \
  8.gene-set-search.ipynb
```

dhimmel:
> I was wondering if you had any insight into why the permuted networks ran so much faster than the original. Shouldn't they be roughly similar?

This is an important observation and I don't know why its occurring. Could the permuted matrix multiplications be less computationally demanding? Perhaps if they produce more sparse DWPC matrices?

kkloste:
> This is an important observation and I don't know why its occurring. Could the permuted matrix multiplications be less computationally demanding? Perhaps if they produce more sparse DWPC matrices?

Permuting the rows or columns of two matrices before multiplying will not change the amount of computation required. This has to be some kind of memory trick --- perhaps permuting the rows of a matrix loads that matrix into cache, so it is faster to multiply in the next step?

PR #78, opened 2018-02-20
Comparison of disk storage methods within NumPy/SciPy

zietzm:
Added a notebook to compare saving and loading HDF5 and NPY sparse vs dense. Saving HDF5 dense took way too long, but it is relatively snappy for sparse matrices, which I read in based on the [compressed-sparse-column format](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csc_matrix.html#scipy.sparse.csc_matrix) of [data, indices, indptr, shape]. In comparing the two, I get very similar results for all except the loading sparse from .npy category. Here, I get either ~2.5 ms or ~12 ms. Will try to nail this down a bit better later.

As it stands, it appears that reading sparse matrices into HDF5 via PyTables is the fastest for reading these matrices back into memory.


pytables.ipynb
```diff
@@ -0,0 +1,287 @@
+{
```
dhimmel: Add title markdown cell at beginning, so viewers know what the point of the notebook is

pytables.ipynb
```diff
@@ -0,0 +1,287 @@
...
+     "text": [
+      "19.55270767211914 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Sparse matrices into .npy\n",
+    "t1 = time.time()\n",
+    "np.save('data/sparse_gig_adj', sparse_gig_adj)\n",
+    "np.save('data/sparse_ctd_adj', sparse_ctd_adj)\n",
```
dhimmel: Are we sure this is saving them as sparse (and not converting them to dense, then saving)? Should we be using [`scipy.sparse.save_npz`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.save_npz.html)?

pytables.ipynb
```diff
@@ -0,0 +1,287 @@
...
+   ],
+   "source": [
+    "t1 = time.time()\n",
+    "h5file = tables.open_file('data/sparse_matrices.h5', 'a')\n",
+    "\n",
+    "node = 'CtD'\n",
+    "data = h5file.get_node(f'/{node[0]}{node[-1]}/{node}/{node}_data').read()\n",
+    "indices = h5file.get_node(f'/{node[0]}{node[-1]}/{node}/{node}_indices').read()\n",
+    "indptr = h5file.get_node(f'/{node[0]}{node[-1]}/{node}/{node}_indptr').read()\n",
+    "mat_shape = h5file.get_node(f'/{node[0]}{node[-1]}/{node}/{node}_shape').read()\n",
```
dhimmel: This is a bit difficult to read. I think that one downside of the HDF5 approach is that it doesn't support sparse matrices as natively.

zietzm: Definitely a downside. The only reason I went about saving so many things was the hope that it could be significantly faster than other formats. If it were much faster, I'd probably just write `pt_read` and `pt_write` functions to make things more smooth with the several different arrays that need to be made for each sparse matrix.

pytables.ipynb
zietzm: I just added a new commit which looks at this. It seems like `np.load` is faster than `sparse.load_npz`, and similarly `np.save` seems faster than `sparse.save_npz`.
One thing I found odd, which perhaps you could offer insight on is the following: when I first run (`Restart & Run All`), I get something like `25 ms` for the sparse load time from `.npz`. When I rerun the cell, the time invariable drops to more like `5 ms`. There must be some kind of caching going on here, right?

kkloste: That does sound fishy, @zietzm , and caching of some sort would be my first thought.

Try running the cell, then running a bunch of other cells (to fill up cache), and running the sparse load cell again, to test the hypothesis?

zietzm: Strange. After working on some other things for a while and running other cells, I retried it and was getting ~ `15 ms`, which wasn't changing much at all. I opened it 100 times and got an average speed of `12.8 ms` for the two matrices.

pytables.ipynb
```diff
@@ -0,0 +1,417 @@
...
+     "output_type": "stream",
+     "text": [
+      "645.2720165252686 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Dense matrices into .npy\n",
+    "t1 = time.time()\n",
+    "np.save('data/gig_adj', gig_adj)\n",
```
dhimmel: Let's add `.npy` to `file` argument. Numpy adds it automatically if missing, but that was a bit unexpected to me. I don't really like the idea of the program changing the file name. Also applies to `.npz` below.

pytables.ipynb
dhimmel: What could explain this?

1. imports. The first time you run the loader, it may have to import some libraries. This only happens for the first time the modules are imported.

2. Swapping. If you're using all your RAM (memory), perhaps you end up having to free up memory or writing to the high-latency swap. These could both be slow.

pytables.ipynb
```diff
@@ -0,0 +1,417 @@
...
+     "text": [
+      "186.43593788146973 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Sparse matrices into .npz with compression\n",
+    "t1 = time.time()\n",
+    "sparse.save_npz('data/sparse_gig_adj', sparse_gig_adj, compressed=True)\n",
+    "sparse.save_npz('data/sparse_ctd_adj', sparse_ctd_adj, compressed=True)\n",
```
dhimmel: To really evaluate whether to compress or not, we'd need to know:

1. how the disk space usage compares
2. how the read-time compares. reading compressed data is sometimes faster, sometimes slower

In our case, minimizing disk space and fast reading are big priorities. However, dense arrays will likely cause most of the problems so this may not be super relevant.

pytables.ipynb
dhimmel: There seems to be an odd naming collision between scipy and numpy. In numpy, `.npy` is for saving a single array and `.npz` stores dictionary of arrays. However, scipy saves single arrays with the `.npz` extension. Confusing.

pytables.ipynb
```diff
@@ -0,0 +1,417 @@
...
+     "text": [
+      "12.098550796508789 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Sparse matrices out of .npy\n",
+    "t1 = time.time()\n",
+    "sparse_gig_adj_load = np.load('data/sparse_gig_adj.npy')\n",
+    "sparse_ctd_adj_load = np.load('data/sparse_ctd_adj.npy')\n",
```
dhimmel: It's confusing for sparse to be in this filename? since it was saved with `numpy.save`?

dhimmel:
Would be nice to compute a DWPC matrix like GiGpBP and see times on that, as well as evaluate what memory mapping (`mmap_mode='r'` in numpy.load) will accomplish. Specifically will it make multiplying a mostly zero 1d array very quick?

kkloste:
> Specifically will it make multiplying a mostly zero 1d array very quick?

My understanding of mmep is that it would not improve runtime for an array of mostly 0s --- mmep seems to be for avoiding loading larger files into RAM in their entirety, instead enabling access to bits and pieces of the large file without loading the whole thing.

If you want to decrease the time spent multiplying arrays with mostly zero-value elements, then sparse type matrices should be the way to go. Sparse matrix data structures organize the matrix content so that the zeros never get accessed at all.

dhimmel:
> mmep seems to be for avoiding loading larger files into RAM in their entirety, instead enabling access to bits and pieces of the large file without loading the whole thing.

Yes that is a desirable feature. Say we have a DWPC matrix precomputed and saved on disk. Then the use provides seed weights for each source node. Most seed weights are zero, thus those rows never need to be read in.

kkloste:
Ah, neat!

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,320 @@
...
+    "    print(f'dense: {dense_size :.4} MB\\nsparse: {sparse_size :.4} MB\\nsparse_uncompressed: {sparse_un_size :.4} MB\\n')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Memory-map reduces matrix-vector multiplication time for lower-density matrices\n",
+    "\n",
+    "Putting a normal load above a memmap load decreases memmap load time. However, loading memmap first does not speed up normal load times. Note, this only appears to be the case for very dense matrices. For best comparisons, we should always load a memmap matrix first in the sequence."
```
dhimmel: It looks to me like the first time you call `np.load` with `mmap_mode='r'`, memory-mapped reading is slower. However, subsequent repeated reads of the same matrix in memory-mapped mode take similar time. This makes me think the slowness of the initial load has to do with importing the memory-map machinery and it's actually subsequent reads that we care about.

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,320 @@
...
+    "    output = vec @ matrix\n",
+    "    t2 = time.time()\n",
+    "    time_1 = t2 - t1\n",
+    "    del matrix\n",
+    "    del output\n",
+    "    \n",
+    "#     Second memory-map load time for comparison\n",
+    "    t1 = time.time()\n",
+    "    matrix = np.load(f'data/{i}.npy', mmap_mode='r')\n",
+    "    output = vec @ matrix\n",
```
dhimmel: It'd be cool to see an additional comparison: Subset matrix for nonzero rows, then multiply. If memmap is very smart, it could be this efficient with the current method. If mmap is dumb, this could speed things up?

It may also affect the times of multiplying non-mmaped matrices.

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,320 @@
...
+   "source": [
+    "np.random.seed(0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sparse .npz (compressed) takes the least disk space regardless of density\n",
+    "Uncompressed sparse matrices take more disk space when the matrices are very dense"
```
dhimmel: I look to source.

When scipy saves sparse matrices, [it uses](https://github.com/scipy/scipy/blob/v1.0.0/scipy/sparse/_matrix_io.py#L76-L79) `numpy.savez` or `numpy.savez_compressed`. It's saving the indices and offsets (arrays) just like yi was in HDF5. This now makes sense why they use .npz extension.

Then we take a look at numpy's `savez` and `savez_compressed` [source. It's using ZIP.](https://github.com/numpy/numpy/blob/v1.14.0/numpy/lib/npyio.py#L689-L698)

Sparse_and_dense.ipynb
zietzm: Not quite sure what you mean. If we subset the matrix after loading, then it would not have correct dimensions for the matrix-vector multiplication.

dhimmel: You can subset the `matrix` and `vec` based on nonzero entries of `vec`... right?

zietzm: Oh I see what you mean: subset based on the nonzero rows, meaning rows corresponding to the search nodes. That makes sense, will try now.

zietzm: You were right about subsetting! It seems that so long as we create a new `np.ones` vector to multiply the subsetted matrix, then we get at least as quick load/multiplication times when compared to the standard memory map.

zietzm:
Upon some suggestions by @dhimmel, I did a comparison of [numpy's memory map methods](https://docs.scipy.org/doc/numpy/reference/generated/numpy.memmap.html#numpy.memmap). As it turns out, we can achieve the fastest load/vector multiplication times when we load a DWPC matrix through memory map, then subset it based on the row indices that correspond to the search nodes. At this size, when I did some comparisons of 4 search nodes, this method was at least as fast as the normal memory map, and at best around 20x faster. 

Using this method, with row indices given, we can return DWPC  results somewhere between 0.05 and 0.00025 seconds, depending on the number of search nodes and the particular matrix in question.

save_matrices.ipynb
```diff
@@ -0,0 +1,82 @@
...
+    "metagraph = graph.metagraph"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in ['CCpGeAeGaD', 'DaGpBP', 'MFpGdCcSE']:\n",
```
dhimmel: Should this match what's in the other notebook:

```
CCpGeAeGaD', 'DaGpBP', 'MFpGdCcSE', 'GiGpBP'
```

***

NVM, I see GiGpBP is done below. Can you add a comment why this metapath is processed separately?

save_matrices.ipynb
```diff
@@ -0,0 +1,82 @@
...
+    "del dwpc_mat"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
dhimmel: Remove empty cell

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,325 @@
...
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "CCpGeAeGaD (1391, 137)\n",
+      "density: 97.8 %\n",
+      "dense: 1.525 MB\n",
+      "sparse: 1.347 MB\n",
+      "sparse_uncompressed: 2.238 MB\n",
```
dhimmel: disambiguating the different storage formats can get a little confusing. Let's be as explicit as possible:

+ dense-uncompressed.npy
+ sparse-compressed.npz
+ sparse-uncompressed.npz

And my suggested file name would be to prepend `{metapath}-`

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,325 @@
...
+   "outputs": [],
+   "source": [
+    "for i in ['CCpGeAeGaD', 'DaGpBP', 'MFpGdCcSE', 'GiGpBP']:\n",
+    "    print(i)\n",
+    "    mat = np.load(f'data/{i}.npy')\n",
+    "    \n",
+    "    density = 100 * (mat != 0).sum() / np.prod(mat.shape)\n",
+    "    print(f'density: {density :.3} %')\n",
+    "\n",
+    "#     Create a vector to multiply\n",
```
dhimmel: Comment should have same indent level in this case.

Sparse_and_dense.ipynb
```diff
@@ -0,0 +1,325 @@
...
+      "sparse_uncompressed: 3.982 MB\n",
+      "\n",
+      "MFpGdCcSE (2884, 5734)\n",
+      "density: 14.1 %\n",
+      "dense: 132.3 MB\n",
+      "sparse: 17.79 MB\n",
+      "sparse_uncompressed: 27.99 MB\n",
+      "\n",
+      "GiGpBP (20945, 11381)\n",
+      "density: 4.65 %\n"
```
dhimmel: Why don't we get file size reports for GiGpBP?

explore/matrix-io/02.sparse-and-dense.ipynb
```diff
@@ -0,0 +1,427 @@
...
+    "    print(f'density: {density :.3} %')\n",
+    "\n",
+    "    sparse_mat = sparse.csc_matrix(mat, copy=True)\n",
+    "\n",
+    "    sparse.save_npz(f'data/{i}-sparse-compressed.npz', sparse_mat, compressed=True)\n",
+    "    sparse.save_npz(f'data/{i}-sparse-uncompressed.npz', sparse_mat, compressed=False)\n",
+    "    \n",
+    "    dense_size = os.path.getsize(f'data/{i}-dense-uncompressed.npy') / 1000000\n",
+    "    sparse_size = os.path.getsize(f'data/{i}-sparse-compressed.npz') / 1000000\n",
+    "    sparse_un_size = os.path.getsize(f'data/{i}-sparse-compressed.npz') / 1000000\n",
```
dhimmel: typo in this line: missing **un**compressed

explore/matrix-io/02.sparse-and-dense.ipynb
```diff
@@ -0,0 +1,427 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "for i in ['CCpGeAeGaD', 'DaGpBP', 'MFpGdCcSE', 'GiGpBP']:\n",
+    "    mat = np.load(f'data/{i}-dense-uncompressed.npy')\n",
+    "    print(i, mat.shape)\n",
+    "    \n",
+    "    density = 100 * (mat != 0).sum() / np.prod(mat.shape)\n",
+    "    print(f'density: {density :.3} %')\n",
```
dhimmel: Python can format percents (no need to multiply by 100) with:

```python
print(f'density: {density:.3%}')
```

explore/matrix-io/02.sparse-and-dense.ipynb
```diff
@@ -0,0 +1,427 @@
...
+    "\n",
+    "    sparse_mat = sparse.csc_matrix(mat, copy=True)\n",
+    "\n",
+    "    sparse.save_npz(f'data/{i}-sparse-compressed.npz', sparse_mat, compressed=True)\n",
+    "    sparse.save_npz(f'data/{i}-sparse-uncompressed.npz', sparse_mat, compressed=False)\n",
+    "    \n",
+    "    dense_size = os.path.getsize(f'data/{i}-dense-uncompressed.npy') / 1000000\n",
+    "    sparse_size = os.path.getsize(f'data/{i}-sparse-compressed.npz') / 1000000\n",
+    "    sparse_un_size = os.path.getsize(f'data/{i}-sparse-compressed.npz') / 1000000\n",
+    "    print(f'dense-uncompressed: {dense_size :.4} MB\\nsparse-compressed: {sparse_size :.4} MB\\nsparse-uncompressed: {sparse_un_size :.4} MB\\n')"
```
dhimmel: Let's change formatter to `.4g`, which should do 4 significant digits and will not do `1.908e+03`.

explore/matrix-io/02.sparse-and-dense.ipynb
```diff
@@ -0,0 +1,427 @@
...
+    "    print(f'Dense No Memmap: {time_1 :.3} sec\\nCompressed Sparse: {time_2 :.3} sec\\nUncompressed Sparse: {time_3 :.3} sec\\n')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true
+   },
+   "source": [
+    "## Subset matrix memmap based on nonzero rows of the search vector"
```
dhimmel: Converting to milliseconds would make it easier to eyeball these times. Another options could be to do time relative to `no memmap` (i.e. no memmap time / memmap time or subset time).

zietzm: I changed that, for the first two sets of cells it will show milliseconds. Then, having established the absolute times, I made the final comparison, subsetting memmap for nonzero rows, compare relative to the dense no-memmap condition. Unfortunately, I am on my laptop until the evening and don't have enough memory to run the notebooks, so they display the old outputs at the moment.

dhimmel: Great excited for the results later!

dhimmel:
> @zietzm I think after we merge, you should open an issue on the numpy repo with a title like "Possible optimization when multiplying memory-mapped by zero". This issue would inquire whether numpy could do the optimization you've demonstrated by subsetting rows. It also could help connect us to the right experts or other users who are doing similar work.

@zietzm do you want to do this or would you like me to?

zietzm:
> do you want to do this or would you like me to?

I'll submit an issue there tonight.

PR #81, opened 2018-03-12
Add pQTL data for pilot analysis

naglem:
Two files with summary data on pQTLs


dhimmel:
Thanks @naglem!

It looks like `data/pqtl_snps_genes_12Mar2018_with_locus_genes_dist.txt` and `data/pqtl_snps_genes_12Mar2018_with_stats.txt` are the same file, if I'm not mistaken.

Some repository style comments:

+ let's move these files to their own directory: `data/pqtl`
+ rename extension from `.txt` to `.tsv` (since they're tab delimited)
+ we can probably simplify the file names a bit? Things like the date can go in the PR comments / commit message.
+ use `-` as word delimiter in file names (unless python modules). For TSV column names, use `_` as word delimiter.

Let me know if you have trouble with any of the following. I'm happy to help if you run into problems.

naglem:
Thanks for noticing that, I must have copied one onto the other unintentionally. I'll update and push them in a second.

naglem:
ok, @dhimmel  new files created and committed. let me know if there is anything else you need, or if there are corrections that are needed.

dhimmel:
For review posting heads of files in https://github.com/greenelab/hetmech/pull/81/commits/d753199f2f0bc95e03c574d85ba7e1bc352378bb below

## `pqtl-snps-genes-with-stats.tsv`

| snp | pqtl_gene | neg_logp | gwas_collection | gwas_source |
|-----------|------------|----------|-----------------|-------------|
| rs651007 | SELE | 102.728 | additional GWAS | 23300549 |
| rs5985 | F13A1,F13B | 185.587 | additional GWAS | 23381943 |
| rs2731672 | F12 | 114.89 | additional GWAS | 23381943 |
| rs10665 | F7 | 46.625 | additional GWAS | 23381943 |
| rs651007 | VWF | 35.355 | additional GWAS | 23381943 |

## `pqtl-snps-genes-with-locus-genes-dist.tsv`

| snp | pqtl_gene | pqtl_entrez_id | locus_gene | locus_entrez_id | distance_to_snp | is_locus_gene_pqtl |
|----------|-----------|----------------|------------|-----------------|-----------------|--------------------|
| rs651007 | SELE | 6401 | HMCN2 | 101060175 | 31079 | FALSE |
| rs651007 | SELE | 6401 | HMCN2 | 256158 | 31079 | FALSE |
| rs651007 | SELE | 6401 | ASS1 | 445 | 41885 | FALSE |
| rs651007 | SELE | 6401 | FUBP3 | 8939 | 175921 | FALSE |
| rs651007 | SELE | 6401 | PRDM12 | 59335 | 261550 | FALSE |
| rs651007 | SELE | 6401 | NCS1 | 23413 | 278848 | FALSE |
| rs651007 | SELE | 6401 | EXOSC2 | 23404 | 290677 | FALSE |

naglem:
I forgot to ask, should I split instances where the protein measured maps to two or more entrez ids? Or can we assess as is?

dhimmel:
For `pqtl-snps-genes-with-stats.tsv`,

+ it'd be best to have standard identifiers for all the `gwas_source`. For example, I think you could change `Biorxiv-Butterworth-2017` to `10.1101/134551` (the DOI of what I think the preprint is). PubMed IDs are also okay for now... just something that leaves no ambiguity.
+ Should `gwas_collection` column be dropped from this public dataset or do we expect it to have more values than just `additional GWAS` in the future?
+ `neg_logp`: confirming this is log not log10? log10 is usually a little nicer for p-values. 
+ `pqtl_gene`: what does it mean when there are two pQTL genes, such as `F13A1,F13B`? Obviously, handing these multi-gene pQTL complicates downstream analyses... so would like a good understanding.
+ consider renaming `pqtl-snps-genes-with-stats.tsv` to `pqtl.tsv`

> I forgot to ask, should I split instances where the protein measured maps to two or more entrez ids? Or can we assess as is?

Just saw the above comment, which relates to my comment. Let's discuss first why there are two pqtl genes.

For `pqtl-snps-genes-with-locus-genes-dist.tsv`, would it be simplest to keep only the following columns:

+ `snp`
+ `locus_gene`
+ `locus_entrez_id`
+ `distance_to_snp`

Then we can `pandas.merge` the two tables as needed. In other words, different studies may report the same SNP twice... but it would only have to appear one time in `pqtl-snps-genes-with-locus-genes-dist.tsv`. If this makes sense, let's rename this file to `snps-to-proximal-genes.tsv` or `snps-to-locus-genes.tsv`.

dhimmel:
One additional todo is to add a README at `data/pqtl/REAMDE.md`. It can be really short. The relevant information is probably a couple sentence description of what the data is and how it was created. In other words:

> pQTL (protein quantitative trait loci) were extracted from public literature by Pfizer. `snps-to-proximal-genes.tsv` provides proximal genes to the SNPs in `pqtl.tsv` according to genome build XXX.

naglem:
Thanks, will alter files to reflect; the reason there could be more than 1 gene ID is that the protein assay measures the protein complex composed of two or more protein coding genes, so in order to map back to entrez ID's, this requires more than one ID. I can flatten this out as well if we think this would be useful.

dhimmel:
> the reason there could be more than 1 gene ID is that the protein assay measures the protein complex composed of two or more protein coding genes

good to know. let's put that in README as well.

> I can flatten this out as well if we think this would be useful.

If you flatten it, then we should make sure to add a column for the protein/complex. If that info is not easily available then I'd say, keep it as is. In short, we want to be able to tell whether rs5985-F13A1-23381943 is from the same association as rs5985-F13B-23381943. Currently, we can tell that, however if you were to flatten, we'd need complex information (or a pQTL ID number to know that).

Flat will be easier for downstream analyses. But I see why intuitively, there should be one row per association (at least until we have a protein complex column). @naglem what're your thoughts?

naglem:
@dhimmel I'd advocate to leave as is for now if you we can analyze as is; as the link is between proteins and the connection is the SNP, as long as we think we can evaluate it in this format then it makes sense to keep.

dhimmel:
> I'd advocate to leave as is for now 

Sounds good.

One final thing I forgot to mention, which will help with tracking in git. It's best to have a deterministic sort order of rows, this way if new rows are added in the future, the diff is minimal. Or if values change, then the line-by-line diffs catch that. The TSVs do look orderly, so I think you probably already considered this.

naglem:
I hadn't, so perhaps I should add a record column to track in the `data/pqtl.tsv` file? Pivoting on the first column could sometimes return different results depending.

dhimmel:
> I hadn't, so perhaps I should add a record column to track in the data/pqtl.tsv file

Not sure what you mean by record column. I would sort `pqtl.tsv` by gwas_source, rs651007, pqtl_gene. The proximity file could be sorted by snp, distance_to_snp.

naglem:
@dhimmel works for me!

dhimmel:
Looks great to me. I think all my comments have been addressed besides the README. Let's also specify in README the max distance for genes listed in the proximity TSV.

naglem:
Done. Should be all set.

Issue #82, opened 2018-03-19
Add pqtl_entrez_id column to pqtl.tsv


dhimmel:
Follows up on https://github.com/greenelab/hetmech/pull/81 by @naglem.

IIRC, some of the in-progress versions of `pqtl.tsv` from https://github.com/greenelab/hetmech/pull/81 had a column titled `pqtl_entrez_id`, which had the genes from `pqtl_gene`, but converted to entrez GeneIDs. It looks like this didn't make it into the merged [`pqtl.tsv`](https://github.com/greenelab/hetmech/blob/e3dd5cdc2b97f235d914d43bf39ff7e41965e482/data/pqtl/pqtl.tsv)

@naglem would you be able to re-add the `pqtl_entrez_id` column? Would make the analysis a tad bit easier.

naglem:
@dhimmel I've updated the data in `data/pqtl/pqtl.tsv` to have a entrez ID column labeled `pqtl_entrez_id`. Let me know if you have any issues with this, thanks.

dhimmel:
Awesome! We'll need to merge these changes in via a pull request. I created https://github.com/greenelab/hetmech/pull/84, but things have gotten a bit messed up since you haven't been creating new branches for each pull request. [Here's a guide](https://github.com/cognoma/cognoma/blob/master/CONTRIBUTING.md#contribution-workflow) on the general pull request workflow. The key is to keep your master branch up to date with `upstream/master` and to start the pull requests from an up-to-date branch. Note that when we merge PRs, we often squash merge so the actual commits end up getting combined into one. This means it's crucial to always pull the commits from `upstream/master`. 

If you create an up-to-date branch, you should be able to apply the single commit that should form the pull request by running:

```
git cherry-pick e65dd8d0e3c434369c442c73415d05ba495d4791
```

naglem:
Hi @dhimmel, not sure if I resolved this, but i created a new branch for this pull request, when I apply the command `git cherry-pick e65dd8d0e3c434369c442c73415d05ba495d4791` it indicates that the branch is up to date, with nothing to commit. Again not sure if it is resolved.

dhimmel:
You will know that things are correct if when you do `git log` you see e65dd8d0e3c434369c442c73415d05ba495d4791 is the most recent commit followed by e3dd5cdc2b97f235d914d43bf39ff7e41965e482.

That's good you know have a branch besides master. Now this is what I think you'll have to do:

```
# get back to master
git checkout master

[...]

PR #84, opened 2018-03-20
Add pqtl_entrez_id column to pqtl.tsv

dhimmel:
Closes https://github.com/greenelab/hetmech/issues/82#issuecomment-374408655


dhimmel:
Closing since the history is messed up as per our discussion in https://github.com/greenelab/hetmech/issues/82#issuecomment-374463741

PR #85, opened 2018-03-20
Add entrez ID column to pqtl.tsv

naglem:
@dhimmel let's see if this works. The branch should be `pqtl_add_col`


dhimmel:
@naglem sadly this PR still has the problem we discuss in https://github.com/greenelab/hetmech/issues/82. Notice how it contains 7 commits. The first 6 commits are from the previous pull request, and were all squashed into one when I merged it. When you've fixed the problem, the pull request will show up as a single commit... just the "added entrez id col" commit.

I do not know how these commits keep sneaking in. The following command was supposed to rewind the history to before they were created:

```
git reset --hard c34d8d94a072418a97ebd14d86412a78fbe17c03
```

I think perhaps you are creating your branch off of a branch with the old incorrect history. Is that possible?

If we can't figure this out soon, I can fix this on my end, but I'd like to get to the point where we fix this and you won't experience this issue again.

PR #87, opened 2018-03-27
Remove duplicate functionality with hetio.matrix

dhimmel:
Migrate calls to hetio.matrix for the following functions:

+ `get_node_to_position`
+ `metaedge_to_adjacency_matrix`
+ `sparsify_or_densify` (previously auto_convert)

Since these functions were moved to hetio as of v0.2.8 in https://github.com/dhimmel/hetio/pull/11


hetmech/matrix.py
```diff
@@ -74,44 +23,20 @@ def normalize(matrix, vector, axis, damping_exponent):
     with numpy.errstate(divide='ignore'):
         vector **= -damping_exponent
     vector[numpy.isinf(vector)] = 0
-    vector = sparse.diags(vector)
+    vector = scipy.sparse.diags(vector)
     if axis == 'rows':
-        # equivalent to `vector @ matrix` but returns sparse.csc not sparse.csr
+        # equivalent to `vector @ matrix` but returns scipy.sparse.csc not scipy.sparse.csr
```
zietzm: This line's length seems to be the only issue I found. Can we make flake8 ignore this particular line, or is it better to change the comment?

dhimmel: In https://github.com/greenelab/hetmech/pull/87/commits/62babfbf1e31004a7ddd90c5a146efb326735ce3 I suppress the line length check for this line. I'm also not against increasing the max line width at a later time. I think 79 is a bit archaic.

PR #89, opened 2018-04-02
Add setup.py

zietzm:
This PR adds a minimal `setup.py` which allows Hetmech to be imported with pip using 
`pip install git+https://github.com/greenelab/hetmech.git@{commit}`

There should be a better way of specifying the install requirements, perhaps using a `requirements.txt` file containing just the necessary components of `environment.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_install:
 install:
   - conda env create --quiet --file environment.yml
   - source activate hetmech
+  - pip install --editable .
 script:
   - pytest hetmech
   - jupyter nbconvert --to=script --stdout *.ipynb |
--- a/README.md
+++ b/README.md
@@ -7,10 +7,6 @@ The method is designed to operate on hetnets (networks with multiple node or rel
 
 This project is still under development. Use with caution.
 
-To use hetmech as a package:
-
-`pip install --editable .`
-
 ## Environment
 
 This repository uses [conda](http://conda.pydata.org/docs/) to manage its environment as specified in [`environment.yml`](environment.yml).
@@ -21,6 +17,10 @@ conda env create --file=environment.yml
 ```
 
 Then use `source activate hetmech` and `source deactivate` to activate or deactivate the environment.
+
+For local development, run the following with the hetmech environment activated:
+
+`pip install --editable .`
 
 ## Acknowledgements
 
--- a/README.md
+++ b/README.md
@@ -3,9 +3,13 @@
 [![Build Status](https://travis-ci.org/greenelab/hetmech.svg?branch=master)](https://travis-ci.org/greenelab/hetmech)
 
 Hetmech aims to identify the relevant network connections between a set of query nodes.
-The method is designed to operate on hetnets (networks with multiple node or relationship types). 
+The method is designed to operate on hetnets (networks with multiple node or relationship types).
 
 This project is still under development. Use with caution.
+
+To use hetmech as a package:
+
+`pip install --editable .`
 
 ## Environment
 
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,5 @@
 from setuptools import setup
 
-from pip.req import parse_requirements
-install_reqs = parse_requirements('environment.yml')
-
 setup(name='hetmech',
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import setup
+
+setup(name='hetmech',
+      description='A search engine for hetnets',
+      long_description='Matrix implementations of path-count-based measures',
+      url='https://github.com/greenelab/hetmech',
+      license='BSD 3-Clause License',
+      packages=['hetmech'],
+      install_requires=['hetio==0.2.8', 'numpy==1.13.1', 'scipy==0.19.1',
+                        'xarray==0.9.6']
```
dhimmel: Let's put each install on its own line. Also remove versions, unless we know we need a min version for a certain package, and then use >=. Whereas environment.yml is meant to define an environment as specifically as possible, setup.py should define it as loosely as possible for greatest compatibility.

dhimmel:
Will the following line automatically update the package when the git repo changes:

```
pip install --editable git://github.com/greenelab/hetmech.git#egg=hetmech
```

I don't think so, because I don't think pip is checking the github version periodically. So perhaps you'd want to do `pip install --editable` within this repo. For the github approach, no benefit of `-e`.

setup.py
zietzm: Makes sense. I left the hetio version in there because we migrated some functions. Do you think it's necessary or could that version label be removed as well?

setup.py
```diff
@@ -0,0 +1,14 @@
...
+      description='A search engine for hetnets',
+      long_description='Matrix implementations of path-count-based measures',
+      url='https://github.com/greenelab/hetmech',
+      license='BSD 3-Clause License',
+      packages=['hetmech'],
+      install_requires=[
+          'hetio>=0.2.8',
+          'numpy',
+          'scipy',
+          'xarray']
```
dhimmel: It's ideal to add `,` then newline before closing bracket. This makes sure all diffs to change or add a new dependency in the future only touch a single line

setup.py
dhimmel: Yep the min version is the right move for hetio since we use functionality that only exists since 0.2.8.

zietzm:
@dhimmel So as it stands, what you can do is imports like the following:
`from hetmech.degree_weight import dwpc`

Is that the behavior we want, or should we add some things in the `__init__.py` file?

dhimmel:
> Is that the behavior we want, or should we add some things in the __init__.py file?

Yeah that behavior is fine. Now user must do:


```
import hetmech.degree_weight
hetmech.degree_weight.dwpc()
```

We could make it so you can do:

```
import hetmech
hetmech.degree_weight.dwpc()
```

Many packages do the later, but I'm not a huge fan. It makes imports more expensive. Some also make functions available in multiple places, like `hetmech.dwpc()` and `hetmech.degree_weight.dwpc()`, which I also don't love because the duplication can be confusing.

dhimmel:
Can you add a `pip install --editable .` to the .travis.yml and the README? Since we're adding a SETUP.md, we should switch to using hetmech everywhere as a package rather than a directory structure.

zietzm:
Just added the line to both, but I may need to move them around within the files.

.travis.yml
```diff
@@ -13,6 +13,7 @@ before_install:
 install:
   - conda env create --quiet --file environment.yml
   - source activate hetmech
+  - pip install --editable .
```
dhimmel: Good

README.md
```diff
@@ -3,10 +3,14 @@
 [![Build Status](https://travis-ci.org/greenelab/hetmech.svg?branch=master)](https://travis-ci.org/greenelab/hetmech)
 
 Hetmech aims to identify the relevant network connections between a set of query nodes.
-The method is designed to operate on hetnets (networks with multiple node or relationship types). 
+The method is designed to operate on hetnets (networks with multiple node or relationship types).
 
 This project is still under development. Use with caution.
 
+To use hetmech as a package:
```
dhimmel: Can we move this to the environment section and say "For local development, run the following with the `hetmech` environment activated:"

Issue #76, opened 2018-02-16
Disk storage of DWPC matrices


zietzm:
Refs #74 

We would like to cache the DWPC matrices on disk, with the ability to rapidly read them. I looked into several different storage methods. HDF5 has a nice Python (somewhat) wrapper called [PyTables](http://www.pytables.org/), which seems useful. In addition, there is a [native Numpy/Scipy saving format,](https://docs.scipy.org/doc/numpy/neps/npy-format.html) .npy/.npz. In a quick notebook example I did earlier today, these two methods seem to read back matrices at comparable speeds (~4.3 ms). With this in mind, I will do some more rigorous speed tests in the next few days, and I will use this issue to organize updates.

dhimmel:
Nice!

> these two methods seem to read back matrices at comparable speeds (~4.3 ms)

That's 233 matrii per second. Not bad. Now what dimension was this matrix? I assume a GxG matrix will take a double-order of mag longer to read in than a DxD.

Some other considerations:

+ we will want to store information on node identity, such as a dataframe of their identifiers and names.  This doesn't necessarily need to be part of the matrix data store, but with HDF5 it could be?

+ [memory-mapping](https://docs.scipy.org/doc/numpy/reference/generated/numpy.memmap.html) could be a big plus in some cases. For example, if someone provides 0-weight for most seed nodes, we don't actually have to read most rows of the matrix (only the rows with nonzero seed weights). How good will each format be for either memory-mapped, subsetted, or piecewise reading?

dhimmel:
See https://github.com/greenelab/hetmech/pull/78 which compares different serialization methods. After this is merged, I think the next step will to get some estimates of total disk space required were we to save everything as `.npy` matrices.

For this we will need to compute the following sets of metapaths for Hetionet v1.0:

1. All metapaths <= length 1
2. All metapaths <= length 2
3. All metapaths <= length 3

and so on, perhaps up to five. Note that currently hetio does not have this feature builtin, although we could add it.

Then we will need to compute the dimensionality of each DWPC matrix and how much disk space it will take up. One optimization may be to log1p the DWPC values allowing us to get by with serializing with a dtype that takes less bytes. @kkloste any insights regarding efficiency saving matrices that consist of small floating point values (usually < 0.01)?

Then we will need to see the total disk space required to store all metapaths for a given length cutoff. Then I think we'll know whether this approach is feasible.

zietzm:
[...]

PR #90, opened 2018-04-09
Scipy sparse vector with memory map

zietzm:
Addresses #76 

Updates [explore/matrix-io/02.sparse-and-dense.ipynb](https://github.com/greenelab/hetmech/blob/master/explore/matrix-io/02.sparse-and-dense.ipynb) to include a comparison of multiplying memory-mapped numpy arrays (which are stored on disk in `.npy` format) by various types of vectors. Specifically, adds a comparison of various `scipy.sparse` matrix formats for this task. 

I found these results fairly surprising, so I reran the notebook several times and found similar results each time. Different types of sparse matrices seem to be better for different tasks, though how this breaks down is a bit tricky. For example, `scipy.sparse.coo_matrix` appears to be the fastest sparse method for multiplication by the largest matrices, while CSR/CSC are basically tied for fastest among the sparse vectors for small matrices. 

Overall, it appears that subsetting memory-mapped matrices and multiplying by a `numpy.ones` vector is still the fastest method in most cases, and **it (subsetting) shows the biggest improvement over other methods for the largest, slowest-to-load matrices.** See [02.sparse-and-dense.ipynb](https://github.com/zietzm/hetmech/blob/1ea4175b7de26c401cdfd7999a1e25b47b7db8bd/explore/matrix-io/02.sparse-and-dense.ipynb)


explore/matrix-io/01.save-matrices.ipynb
```diff
@@ -18,11 +18,17 @@
...
+     "text": [
+      "/bin/sh: 1: {sys.executable}: not found\r\n"
+     ]
+    }
+   ],
    "source": [
-    "import sys\n",
-    "sys.path.insert(0,'../../')\n",
-    "from hetmech.degree_weight import dwpc"
+    "!{sys.executable} -m pip install git+https://github.com/dhimmel/hetio@aa5d12ec43d2dfd349893df9efd9ef8f665c982b --upgrade"
```
dhimmel: Hmm. This is not the right place to upgrade hetio.

zietzm: Right. I was not sure if I should update the `environment.yml` file or if a new hetio release was planned.

dhimmel: Let's update `environment.yml` to use a GitHub commit hash rather than PyPI version. However, this PR doesn't require hetio commits past v0.2.8, so do we want to make the environment change when we need it for the next PR?

zietzm: Ah right. I forgot that I had only added that to a different notebook because I was needing the function `extract_all_metapaths`, but we are not using that here.

PR #93, opened 2018-04-11
Suggest candidate causal genes for each pQTL

dhimmel:
Uses DWPC z-scores for the GpBPpG metapath.

This is more of a proof of concept than complete analysis as eventually we'll want to look at multiple metapaths.


data/pqtl/02.analyze-GpBPpG.ipynb
```diff
@@ -0,0 +1,2043 @@
...
+    "        DWPCs[name] = matrix.multiply(dwpc_scaler ** -1).arcsinh()\n",
+    "    else:\n",
+    "        DWPCs[name] = numpy.arcsinh(matrix / dwpc_scaler)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Unused code"
```
zietzm: Do we want to keep this?

dhimmel: I kept it because we may want to revisit that alternative implementation in the future. Currently, we don't compute full rdwpc and zdwpc matrices (because that code was really slow), but we will probably need to eventually.

PR #94, opened 2018-05-09
Add dtype functionality in DWPC computation

zietzm:


zietzm:
Turns out this is already an issue there.
https://github.com/scipy/scipy/issues/7408

PR #96, opened 2018-05-14
Degree-grouped permtuations

zietzm:
Combine information from different permuted networks and multiple metapaths. One of the goals is that we will have more permuted samples. This should give a greater chance of getting nonzero permuted DWPCs.


zietzm:
Turns out that using degree-grouped permutations reduces the number of zero P-DWPCs by over 99%.

zietzm:
With proper scaling as you describe, there is still a lot of variability between metapaths.

CtDrD
image
image

zietzm:
I will merge this for now, as its purpose is more a proof of concept than a serious implementation. I think it will be necessary to update this notebook once #100 is merged and we have a more developed query pipeline for the search engine.

PR #99, opened 2018-05-16
Update acknowledgements with Pfizer support

dhimmel:
@cgreene and @naglem: I realized we hadn't updated the README to acknowledge Pfizer's support. Is there a specific wording or grant number that I should use to acknowledge this grant?


naglem:
I'll have to check on that, and where the funding sources actually came from internally.

PR #97, opened 2018-05-15
The hetmat awakens: a directory-structure and module for matrix-storage of hetnets

dhimmel:

PR #100, opened 2018-05-17
HetMat permutations: export hetio.hetnet graphs to hetmats

dhimmel:
Follows up on #97.

CC @zietzm


hetmech/hetmat.py
```diff
@@ -39,6 +40,32 @@ def hetmat_from_graph(graph, path):
     return hetmat
 
 
+def hetmat_from_permuted_graph(hetmat, permutation_id, permuted_graph):
+    """
+    Assumes subdirectory structure and that permutations inherit nodes but not
+    edges.
+    """
+    if '.' in permutation_id:
+        raise ValueError('Permutations ids cannot contain `.`')
```
zietzm: Is this a problem because directories cannot have a period in their name? Are ids usually the names like `hetionet-v1.0-perm-1`?

hetmech/hetmat.py
```diff
@@ -39,6 +40,32 @@ def hetmat_from_graph(graph, path):
...
+    if not hetmat.permutations_directory.is_dir():
+        hetmat.permutations_directory.mkdir()
+    directory = hetmat.permutations_directory.joinpath(f'{permutation_id}.hetmat')
+    if directory.is_dir():
+        # If directory exists, back it up using a .bak extension
+        backup_directory = directory.with_name(directory.name + '.bak')
```
zietzm: Why backup existing directory by default?

hetmech/hetmat.py
dhimmel: periods now allowed by https://github.com/greenelab/hetmech/pull/100/commits/ca006c046eef8295a044b8a2bb24ffa27958f984. The implementation is not much uglier, so the ability to support `.` makes sense.

hetmech/hetmat.py
dhimmel: I was thinking that there could be substantial computing power cached in these directories and I didn't want someone accidentally overwritting everything.

PR #102, opened 2018-05-18
degree_weight._dwpc_disjoint: recurse dwpc call for simplicity

zietzm:
Refs #76 

Adds simple `functools.lru_cache`-ing and a helper DWPC function to make this somewhat more optimized.


dhimmel:
I renamed for the mergeable change. Can you remove `disk-cache.ipynb`

PR #104, opened 2018-05-18
Add graph permutations as HetMat method

zietzm:


hetmech/hetmat.py
```diff
@@ -113,6 +114,30 @@ def read_first_matrix(specs):
         '\n'.join(paths))
 
 
+def _permute_matrix(adjacency_matrix, directed=False, multiplier=10,
+                    excluded_pair_set=set(), seed=0, log=False):
+
+    nonzeros = [list(i) for i in adjacency_matrix.nonzero()]
+    edge_list = list(zip(*nonzeros))
```
dhimmel: Did it not work to just do:

```
list(zip(*adjacency_matrix.nonzero()))
```

i.e. not sure why we need to convert arrays to lists in line above, since we're just iterating through them anyways.

hetmech/hetmat.py
```diff
@@ -113,6 +114,30 @@ def read_first_matrix(specs):
...
+    edges = numpy.array(permuted_edges)
+    ones = numpy.ones(len(edges), dtype=adjacency_matrix.dtype)
+    permuted_adjacency = scipy.sparse.csc_matrix((ones, (edges[:, 0], edges[:, 1])),
+                                                 shape=adjacency_matrix.shape)
+
+    # Keep the same sparse type as adjacency_matrix
+    permuted_adjacency = type(adjacency_matrix)(permuted_adjacency)
+
+    # Ensure node degrees have been preserved
+    assert (permuted_adjacency.sum(axis=1) != adjacency_matrix.sum(axis=1)).sum() == 0
```
dhimmel: let's use all:

```python
all(permuted_adjacency.sum(axis=1) == adjacency_matrix.sum(axis=1))
```

hetmech/hetmat.py
```diff
@@ -178,6 +203,57 @@ def permutations(self):
...
+        multiplier : int
+        excluded_pair_set : set
+        seed : int
+        log : bool
+        """
+        if namer is None:
+            # If no namer given, continue increasing names by one for new permutations
+            def namer():
+                i = 1
+                while i:
```
dhimmel: `while True` is the convention here.

hetmech/hetmat.py
```diff
@@ -178,6 +203,57 @@ def permutations(self):
...
+            if start_from is not None:
+                original_hetmat = self.permutations[start_from]
+            else:
+                original_hetmat = self
+
+            metaedges = list(self.metagraph.get_edges(exclude_inverts=True))
+            for metaedge in metaedges:
+                rows, cols, original_matrix = original_hetmat.metaedge_to_adjacency_matrix(
+                    metaedge, dense_threshold=1)
+                is_directed = (metaedge.direction != 'both')
```
dhimmel: uneeded parentheses

hetmech/hetmat.py
```diff
@@ -178,6 +203,57 @@ def permutations(self):
...
+                permuted_matrix, stats = _permute_matrix(
+                    original_matrix, directed=is_directed, multiplier=multiplier,
+                    excluded_pair_set=excluded_pair_set, seed=seed, log=log)
+                path = new_hetmat.get_edges_path(metaedge, file_format='sparse.npz')
+                scipy.sparse.save_npz(str(path), permuted_matrix, compressed=True)
```
dhimmel: Could there be a potential issue where `_permute_matrix` doesn't return something sparse?

dhimmel: Isn't this converting the permutated matrix to the type of the input?

```diff
+    # Keep the same sparse type as adjacency_matrix
+    permuted_adjacency = type(adjacency_matrix)(permuted_adjacency)
```

zietzm: That only allows the interconversion between different types of sparse matrices. Do we want to be able to save them either way?

dhimmel: Not necessarily, but will this break if someone passes a dense array to the function?

zietzm: Just added dense support

hetmech/hetmat.py
```diff
@@ -197,6 +224,71 @@ def permutations(self):
...
+                rows, cols, original_matrix = original_hetmat.metaedge_to_adjacency_matrix(
+                    metaedge, dense_threshold=1)
+                is_directed = metaedge.direction != 'both'
+                permuted_matrix, stats = _permute_matrix(
+                    original_matrix, directed=is_directed, multiplier=multiplier,
+                    excluded_pair_set=excluded_pair_set, seed=seed, log=log)
+                path = new_hetmat.get_edges_path(metaedge, file_format=None)
+                save_matrix(permuted_matrix, path)
+                logging_stats.append(stats)
+            start_from = permutation_name
```
dhimmel: This does not seem neccessary at this point?

hetmech/hetmat.py
```diff
@@ -197,6 +224,71 @@ def permutations(self):
...
+            name = f'{permutation_name}.hetmat/'
+            new_permutation_path = self.permutations_directory.joinpath(name)
+
+            new_hetmat = HetMat(new_permutation_path, initialize=True)
+            new_hetmat.is_permutation = True
+            new_hetmat.metagraph_path.symlink_to('../../metagraph.json')
+            new_hetmat.nodes_directory.rmdir()
+            new_hetmat.nodes_directory.symlink_to('../../nodes', target_is_directory=True)
+
+            if start_from is not None:
```
dhimmel: Would be possible to simplify by renaming `original_hetmat` to `start_from`. User could then pass `None` for self, a hetmat for a specific loaded hetmat, or a str for the name of a permtuation.

hetmech/hetmat.py
```diff
@@ -197,6 +224,77 @@ def permutations(self):
...
+        excluded_pair_set : set
+            Pairs of nodes which are not to be changed
+        seed : int
+            Random seed for generating new permutations
+        log : bool
+            When log=True, give information about permutation attempts
+        """
+        if namer is None:
+            # If no namer given, continue increasing names by one for new permutations
+            def namer():
```
dhimmel: replace with:

```python
namer = itertools.count(start=1)
```

hetmech/hetmat.py
```diff
@@ -197,6 +224,77 @@ def permutations(self):
...
+                i = 1
+                while True:
+                    yield str(i).zfill(3)
+                    i += 1
+            namer = namer()
+
+        logging_stats = list()
+        for _ in range(num_new_permutations):
+            permutation_name = next(namer)
+            name = f'{permutation_name}.hetmat/'
```
dhimmel: trailing slack no good?

hetmech/hetmat.py
zietzm: If not, then we need to increment the seed after each iteration. Otherwise would we not be outputting multiple copies of the same permutation?

dhimmel: Disregard this comment... I was misunderstanding

hetmech/matrix.py
```diff
@@ -66,3 +67,29 @@ def copy_array(matrix, copy=True, dtype=numpy.float64):
         mat_type = numpy.array
     matrix = mat_type(matrix, dtype=dtype, copy=copy)
     return matrix
+
+
+def permute_matrix(adjacency_matrix, directed=False, multiplier=10,
+                   excluded_pair_set=set(), seed=0):
+
```
dhimmel: Let's add a docstring that this currently assumes boolean adjacency matrices... i.e. it doesn't deal with edge weigts

PR #107, opened 2018-05-22
Generate new permuted HetMats

zietzm:


dhimmel:
Let's add `float_format='%.5g'` to the to_csv command, as we did [here](https://github.com/dhimmel/integrate/blob/93feba1765fbcd76fd79e22f25121f5399629148/permute.ipynb)

zietzm:
Quick note: https://github.com/greenelab/hetmech/pull/107/commits/5c58e401a7c71da03360547e53115819075d9a99 did not show changes in any of the permutations themselves. This shows that the permutation method is replicable and deterministic.

Issue #108, opened 2018-05-22
Caching and bulk computation of metapaths


zietzm:
If we are doing bulk computations of metapaths, we will want to take full advantage of our developing caching functionality. During the computation of most DWPCs, the metapath will be segmented and sub-problems computed first. We do not want to cache metapaths or metapath segments that appear only once in the full list of segments. Excluding these single-appearance segments **excludes 78% of all segments** (regardless of whether looking at `max_length=4` or `max_length=5`).

Another thing to think about could be retrieval vs transpose time for some segments. For example, the first and sixth most common segments are GdD and DdG. If transposing is much slower than retrieving, maybe we want to cache both the GdD and DdG matrices. But for sparse matrices, transposing may be quite quick, in which case we would prefer storing one direction only.

I have rank-ordered all segments by their frequency of appearance. An interesting thing to note is that nearly one-third of all segments which would be included (meaning appear more than once) appear less than ten times, while only ten percent appear more than one hundred times. **Basically, the distribution of frequencies is extremely biased toward small numbers with an incredibly long tail**. The most frequent segments occur nearly 30,000 times, while only 1% occur even 1000 times. Thus, the smallest increases in performance will potentially be quite noticeable for the most frequent segments, while others may be much less visible.

dhimmel:
I think this issue has mostly been addressed through the following pull requests:

- https://github.com/greenelab/hetmech/pull/109
- https://github.com/greenelab/hetmech/pull/111
- https://github.com/greenelab/hetmech/pull/112

One thing on the wishlist is a cache method to print stats. We can leave this issue open a bit longer in case to discuss any lingering cache issues.

dhimmel:
Will close since we've implemented caching... although perhaps imperfectly as there seems to be a severe memory leak: https://github.com/greenelab/hetmech/issues/141.

PR #109, opened 2018-05-24
Prioritize segments for caching

zietzm:
Relevant to #108 

When doing bulk computations of DWPC, we want to cache the metapath segments that occur most frequently. To find these, we use two new functions, `get_all_segments` and `order_segments`.

`get_all_segments` returns a list of all segments that will be found from continued metapath splitting (to completion). An important note here is that I decided keeping segments from each split would be best. For example, given the following metapath: `CrCbGaDrDaG`, we could have gotten all segments in two ways.

Rather than returning:
[`CrC`, `CbG`, `GaD`, `DrD`, `DaG`],

I think it is more helpful to return
[`CrC`, `CbG`, `GaD`, `DrD`, `DaG`, `GaDrDaG`].

This way, if the added segment appears very frequently over all the metapaths for which we want to compute DWPC, then we would prefer to cache the larger segment itself rather than only the smallest units. This way of splitting allows us to make these decisions automatically.


dhimmel:
> This way, if the added segment appears very frequently over all the metapaths for which we want to compute DWPC

Okay so this approach is going to return exactly the same number of segments as the number of times the `dwpc` function is called?

hetmech/degree_weight.py
```diff
@@ -368,6 +368,75 @@ def add_head_tail(metapath, indices):
...
+        Whether or not to include both forward and backward directions of segments.
+        For example, if False: [CbG, GbC] -> [CbG, CbG], else no change.
+
+    Returns
+    -------
+    collections.Counter
+        Number of times each metapath segment appears when getting all segments.
+    """
+    all_segments = [segment for metapath in metapaths for segment in get_all_segments(metagraph, metapath)]
+    if not store_inverses:
```
dhimmel: I think your method works although it's a bit challenging to wrap my mind around. Here's the approach I'd take:

```python
seen = set()
aligned_segments = list()
for segment in all_segments:
    add = segment.inverse if segment.inverse in seen else segment:
    aligned_segments.append(add)
    seen.add(add)
```

hetmech/degree_weight.py
```diff
@@ -368,6 +368,75 @@ def add_head_tail(metapath, indices):
...
+        stored_directions = set(all_segments)
+        for i, v in enumerate(all_segments):
+            if v.inverse in stored_directions:
+                if v in stored_directions:
+                    stored_directions.remove(v.inverse)
+                else:
+                    all_segments[i] = v.inverse
+    segment_counts = collections.Counter(all_segments)
+    # Keep only those which appear more than once
+    segment_counts = collections.Counter({k: v for k, v in segment_counts.items() if v > 1})
```
dhimmel: Let's keep single-count segments, since I think we may want to compute metrics that would like to know this number... i.e. cacheability. However, it's a good insight that we would not want to cache these, however that exclusion can occur in the caching class.

hetmech/degree_weight.py
```diff
@@ -368,6 +368,75 @@ def add_head_tail(metapath, indices):
...
+
+    Example
+    -------
+    >>> get_all_segments(metagraph, CrCbGaDrDaG)
+    [CrC, CbG, GaDrDaG, GaD, DrD, DaG]
+    """
+    segments = get_segments(metagraph, metapath)
+    if len(segments) == 1:
+        return [metapath]
+    else:
```
dhimmel: this else can be zapped as return shortcircuits

hetmech/degree_weight.py
zietzm: Fixed!

zietzm:
@dhimmel I don't think that I can make the visualization that you suggest without deciding on a caching scheme. Have we decided how many results, or what total size to cache?

hetmech/degree_weight.py
```diff
@@ -368,6 +368,71 @@ def add_head_tail(metapath, indices):
...
+    """
+    all_segments = [segment for metapath in metapaths for segment in get_all_segments(metagraph, metapath)]
+    if not store_inverses:
+        # Change all instances of inverted segments to the same direction, using a first-seen ordering
+        seen = set()
+        aligned_segments = list()
+        for segment in all_segments:
+            add = segment.inverse if segment.inverse in seen else segment
+            aligned_segments.append(add)
+            seen.add(add)
```
dhimmel: At the end I think you need `all_segments = aligned_segments`

zietzm:
image
Here's the plot. Looks very promising for a very limited cache.

PR #112, opened 2018-05-29
DWWC recursive and chain-ordering

zietzm:
References #53 

Add two new functions for computing DWWC. The notebook shows that both chain ordering and recursion with caching are a good bit faster than our standard implementation. Note that in the notebook, I performed the computations having `dwwc_recursive` cache using the cacher metric `dwpc`. This was to make sure that `dwwc` was not simply calling the cached result of the recursive function, but to include the actual times we would expect when using caching.

Over the Rephetio metapaths, the two new methods took this many seconds (old method for comparison) in sparse format.

| DWWC method | Time for all Rephetio (s) |
|-------------|---------------------------|
| original    | 219.110172                |
| recursive   | 93.744713                 |
| chain       | 70.896228                 |


zietzm:
I just added another chain multiplication function, this time which does not keep all the arrays in memory at the same time, as was mentioned in #53. The two functions appear to have fairly similar performance, with the `dwwc_chain_nomem` function a tiny bit faster. We can choose however we prefer to do this.

hetmech/degree_weight.py
```diff
@@ -137,6 +137,7 @@ def dwwc(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
         converted to a dense automatically.
     dtype : dtype object
     """
+    metapath = graph.metagraph.get_metapath(metapath)
```
dhimmel: We shouldn't need this because the decorator should already be doing this conversion?

hetmech/degree_weight.py
```diff
@@ -152,6 +153,116 @@ def dwwc(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
...
+@path_count_cache(metric='dwwc')
+def dwwc_recursive(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
+    """
+    Recursive DWWC implementation to take better advantage of caching.
+    """
+    metapath = graph.metagraph.get_metapath(metapath)
+    rows, cols, adj_mat = hetmech.matrix.metaedge_to_adjacency_matrix(
+        graph, metapath[0], dense_threshold=dense_threshold, dtype=dtype)
+    adj_mat = _degree_weight(adj_mat, damping, dtype=dtype)
+    if len(metapath[1:]) > 0:
```
dhimmel: would `if len(metapath) > 1` be simpler?

hetmech/degree_weight.py
```diff
@@ -152,6 +153,116 @@ def dwwc(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
...
+        Actually do the multiplication with the given order. Modified from
+        original at https://git.io/vh31f
+        """
+        if i == j:
+            _, _, adj_mat = metaedge_to_adjacency_matrix(
+                graph, metapath[i], dense_threshold=dense_threshold, dtype=dtype)
+            adj_mat = _degree_weight(adj_mat, damping)
+            return adj_mat
+        else:
+            return numpy.dot(_multi_dot(metapath, order, i, order[i, j]),
```
dhimmel: Does this work for sparse matrices?

zietzm: It appears to work, but I am switching it to use `@`

hetmech/degree_weight.py
zietzm: Good call. Do you think we will want to put the decorator on the chain function as well?

zietzm:
You are absolutely right, I had forgotten to configure the cache! I will rerun that and compare.

Quick note of interest: what appears to be the fastest dwwc method, `dwwc_chain_nomem`, computed DWWC for all metapaths length <= 4 in almost exactly 11 hours. Will rerun this evening using the recursive function with a properly set up cache.

zietzm:
So looking through the runtimes, `dwwc_recursive` is slower than `dwwc` on about 20% of metapaths. Over the Rephetio paths, though, `dwwc_recursive` is clearly much faster, taking 90 rather than 225 seconds to compute all DWWCs. I think the reason `dwwc_recursive` is faster on some metapaths is that, if not using a cache, it keeps more matrices in memory at once. So while the peak memory usage is pretty close for computations using `dwwc` and `dwwc_recursive`, the peak memory usage is slightly higher ultimately for `dwwc_recursive`.

In computing other metapath DWWCs, `dwwc_recursive` is slower. Here are the top four metapaths with the biggest time difference in favor of `dwwc`, ie. those in which it was most an improvement over `dwwc_recursive`. Nothing really stands out to me with these metapaths. @dhimmel perhaps you can notice something with these that would explain the difference in computation times. I suppose there could also be some element of randomness here, as these times are all quite small.

image

zietzm:
It appears not to make a big difference which DWWC method is used for DWPC computation. In the notebook I just added, `which-dwpc.ipynb`, the time required to compute all Rephetio paths was roughly equivalent between all three methods, with `dwwc_chain` being slightly faster than the other two.

explore/dwwc-chain-recursive.ipynb
```diff
@@ -0,0 +1,702 @@
...
+    "df = pd.DataFrame(all_times, columns=['metapath', 'abbrev', 'dwwc', 'dwwc-nocache', 'recursive', 'recursive-nocache', 'chain', 'chain-nocache'])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df.to_csv('../data/rephetio-DWWCs-hetmat-runtime.tsv', sep='\\t')"
```
dhimmel: `index=False` and `float_format='%.5g'`

explore/dwwc-chain-recursive.ipynb
```diff
@@ -0,0 +1,702 @@
...
+       "      <td>0.019714</td>\n",
+       "      <td>0.019020</td>\n",
+       "      <td>0.018035</td>\n",
+       "      <td>0.020231</td>\n",
+       "    </tr>\n",
+       "  </tbody>\n",
+       "</table>\n",
+       "</div>"
+      ],
+      "text/plain": [
```
dhimmel: Interesting, when written to TSV, both metapath and abbrev show the abbreviation.

dhimmel:
It would be nice to see if we can figure out why recursive-nocache is so much faster than dwwc-nocache (which is equivalent here to dwwc since dwwc is not recursive and has no submetapath caching). Using a profiler like [line_profiler](https://github.com/rkern/line_profiler) could help.

hetmech/degree_weight.py
```diff
@@ -152,6 +152,75 @@ def dwwc(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
...
+    if i == j:
+        _, _, adj_mat = metaedge_to_adjacency_matrix(
+            graph, metapath[i], dense_threshold=dense_threshold, dtype=dtype)
+        adj_mat = _degree_weight(adj_mat, damping=damping)
+        return adj_mat
+    return _multi_dot(metapath, order, i, order[i, j], graph, damping, dense_threshold, dtype) \
+        @ _multi_dot(metapath, order, order[i, j] + 1, j, graph, damping, dense_threshold, dtype)
+
+
+def dwwc_chain(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
```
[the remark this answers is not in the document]
zietzm: In the notebooks it showed no improvement. However, I guess that just like `dwwc` it should probably be there anyway in cases of multiple identical calls.

hetmech/degree_weight.py
```diff
@@ -677,12 +746,12 @@ def _dwpc_short_repeat(graph, metapath, damping=0.5, dense_threshold=0,
     col_names = cols
 
     if head_segment:
-        row_names, cols, head_dwpc = dwwc(graph, head_segment, damping=damping,
+        row_names, cols, head_dwpc = dwpc(graph, head_segment, damping=damping,
```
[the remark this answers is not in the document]
zietzm: This was just an instance that I missed earlier. Basically, just making sure that all segments go back through the original function. Doesn't make a big difference either way.

zietzm:
Just pushed one more commit. Will rerun notebooks (may take a while) and then this PR should be done.

hetmech/degree_weight.py
```diff
@@ -152,6 +152,80 @@ def dwwc(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
...
+                    m[i, j] = q
+                    ordering[i, j] = k
+    return ordering
+
+
+@path_count_cache(metric='dwwc')
+def dwwc_chain(graph, metapath, damping=0.5, dense_threshold=0, dtype=numpy.float64):
+    """
+    Uses optimal matrix chain multiplication as in numpy.multi_dot, but allows
+    for sparse matrices. See original source, numpy.linalg: https://git.io/vh38o
```
dhimmel: Must also mention license of original here.

zietzm:
Closed #53

PR #114, opened 2018-06-05
Degree-grouped permutations implementation

zietzm:


dhimmel:
Must fix LFS... probably should move .gitattributes to root

dhimmel:
We probably want to output each metapath as a separate TSV. This will be more maneagable.

dhimmel:
We do not want to filter zero-DWPC rows.

.gitignore
```diff
@@ -1,5 +1,6 @@
 /hetnet/adjacency/data-big
 /data/hetionet-v1.0.hetmat/permutations/*.bak
+/data/hetionet-v1.0.hetmat/permutations
```
dhimmel: We are currently tracking permutations?

PR #115, opened 2018-06-06
Add get_stats method to cache objects

dhimmel:
Refs https://github.com/greenelab/hetmech/issues/108#issuecomment-395075161

### DIFF
--- a/explore/cache-speeds.ipynb
+++ b/explore/cache-speeds.ipynb
@@ -55,6 +55,7 @@
     "    if hetmech.degree_weight.categorize(metapath) in {'long_repeat', 'other'}:\n",
     "        continue\n",
     "    metapaths.append(metapath)\n",
+    "# metapaths = metapaths[:3]  # for development\n",
     "len(metapaths)"
    ]
   },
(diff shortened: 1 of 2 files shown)


dhimmel:
Will add another commit with updates to `explore/cache-speeds.ipynb`.

hetmech/hetmat.py
```diff
@@ -442,12 +443,24 @@ def set(self, metapath, metric, damping, matrix, runtime):
...
+              cache hits: {hits_str}''')
+        return stats_str
+
 
 class PathCountPriorityCache(PathCountCache):
 
     def __init__(self, hetmat, allocate_GB):
         super().__init__(hetmat)
-        self.allocate_B = 1_000_000_000 * allocate_GB
+        self.bytes_per_gigabyte = 1_000_000_000
```
zietzm: Why do we want this as an attribute, when `self.bytes_per_gigabyte` is longer than the number itself or `1e9`, which are equivalent?

dhimmel: 1e9 creates float. I felt that there is less of a change to get an error if we only set the conversion ratio once and it's semantically more expressive.

explore/cache-speeds.ipynb
```diff
@@ -55,6 +55,7 @@
     "    if hetmech.degree_weight.categorize(metapath) in {'long_repeat', 'other'}:\n",
     "        continue\n",
     "    metapaths.append(metapath)\n",
+    "# metapaths = metapaths[:3]  # for development\n",
```
zietzm: Maybe want to remove this line once done.

PR #116, opened 2018-06-06
Add support for four-repeats.

zietzm:
Eg: `GiGiGiG`


hetmech/degree_weight.py
```diff
@@ -853,6 +869,7 @@ def _dwpc_general_case(graph, metapath, damping=0, dtype=numpy.float64):
     damping : float
     dtype : dtype object
     """
+    metapath = graph.metagraph.get_metapath(metapath)
```
dhimmel: This is not necessary right, because it's handled at a higher stage?

PR #117, opened 2018-06-06
Add DWPC approximation method

zietzm:
My approximation method is basically the following:

1. Find the first repeated metanode.
2. Compute DWPC in the segment between each occurrence of this metanode, removing the diagonal afterwards.
3. Compute DWPC for the head and tail segments if they are present.

For example:

Given CbGiGaDrDaGpBP
1. G, [1,2,5]
2. `remove_diag(GiG @ dwpc(GaDrDaG))`
3. CbG @ GiGaDrDaG @ GpBP

I have changed so that under `use_general=False`, rather than returning an error, we simply output an approximate.

### DIFF
--- a/hetmech/degree_weight.py
+++ b/hetmech/degree_weight.py
@@ -116,6 +116,9 @@ def dwpc(graph, metapath, damping=0.5, dense_threshold=0, approx_ok=False,
     """
     category = categorize(metapath)
     dwpc_function = _category_to_function(category, dwwc_method=dwwc_method, approx_ok=approx_ok)
+    if category in ('long_repeat', 'other') and not approx_ok:
+        logging.warning(f"Metapath {metapath} will use _dwpc_general_case, "
+                        "which can require very long computations.")
     row_names, col_names, dwpc_matrix = dwpc_function(
         graph, metapath, damping, dense_threshold=dense_threshold,
         dtype=dtype)
--- a/hetmech/degree_weight.py
+++ b/hetmech/degree_weight.py
@@ -20,7 +20,7 @@ from hetmech.matrix import (
 )
 
 
-def _category_to_function(category, dwwc_method, approx_ok):
+def _category_to_function(category, dwwc_method):
     function_dictionary = {
         'no_repeats': dwwc_method,
         'disjoint': _dwpc_disjoint,
@@ -34,9 +34,6 @@ def _category_to_function(category, dwwc_method, approx_ok):
         'interior_complete_group': _dwpc_baba,
         'other': _dwpc_approx,
     }
-    if approx_ok:
-        function_dictionary.update({'long_repeat': _dwpc_general_case,
-                                    'other': _dwpc_general_case})
     return function_dictionary[category]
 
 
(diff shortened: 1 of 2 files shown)


dhimmel:
@zietzm is this the method metioned by @mmayers12 in https://github.com/greenelab/hetmech/issues/58#issuecomment-314516251:

> So now we have a method that overestimates the error and a method that underestimates it. In order to get a fast approximation for DWPC in these cases, we take an element-wise average of the two errors matrices (sum and element-wise max) to get something that's fairly close to the true DWPC while still calculating the result in under 5 seconds.

@mmayers12 has your team had any more experience or insights with the approximation method since then? We have exact methods for many metapaths, so this approximation method is meant to fill in the gaps.

zietzm:
This method is somewhat similar. Basically, for an arbitrarily complicated metapath, this method will search for the first metanode which is repeated more than once, and perform repeated multiplications/diagonal subtractions. It's worth noting that this particular method does not need to perfectly handle many metapath types, (such as BABA, ABBA, etc.) because they will already be handled by our other, exact methods, though this does handle any such regions within other repeats.

For a metapath like **A**.B.C.**A**.B.C.D.**A**, we do basically,

`remove_diag(remove_diag(dwpc(ABCA)) * dwpc(ABCDA))`

This means that even cases like **A**.B.B.B.**A**.C.C.C.**A** would have exact results. Only more complicated overlaps will not be exact, for example, **A**.**A**.B.C.**A**.**A**.B.C

hetmech/degree_weight.py
```diff
@@ -113,19 +115,13 @@ def dwpc(graph, metapath, damping=0.5, dense_threshold=0, use_general=False,
...
-        else:
-            raise NotImplementedError(
-                'Metapath category will use _dwpc_general_case')
-    else:
-        row_names, col_names, dwpc_matrix = dwpc_function(
-            graph, metapath, damping, dense_threshold=dense_threshold,
-            dtype=dtype)
-
+    dwpc_function = _category_to_function(category, dwwc_method=dwwc_method, approx_ok=approx_ok)
+    if category in ('long_repeat', 'other') and not approx_ok:
```
dhimmel: One option is to leave `_category_to_function` alone, and just do:

```python
if category in ('long_repeat', 'other'):
    if approx_ok:
        dwpc_function = _dwpc_approx
    else:
        # Do warning
```

hetmech/tests/test_degree_weight.py
```diff
@@ -636,3 +636,26 @@ def test_dtype(metapath, dtype, dwwc_method):
...
     assert dwpc_matrix.dtype == dtype
+
+
+@pytest.mark.parametrize('metapath,relative', [
+    ('DrDaGiG', 'equal'),
+    ('DaGiGaD', 'equal'),
+    ('DaGaDrDaGaD', 'not_equal'),
+    ('CrCpDrD', 'equal')
+])
+def test_dwpc_approx(metapath, relative):
```
dhimmel: Could there be an issue if we change the dense_threshold default of dwpc in the future? Or would these tests work for sparse and dense?

zietzm: These comparisons work either way.

PR #118, opened 2018-06-11
Compare Z-DWPC distributions

zietzm:
I think the [Kolmogorov–Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) is a good way to compare the distribution of observed (calculated) Z-DWPC values to the expected distribution of Z-values (the standard normal distribution). In the notebook is an implementation of this metric for the comparison of Z-DWPC scores.


zietzm:
img-2616

In essence, this test will calculate the area of the shaded region, as a way to quantify how many more high Z-values we observe than we would expect.

hetmech/degree_group.py
```diff
@@ -72,13 +70,11 @@ def compute_summary_metrics(df):
 
 def dwpc_to_degrees(graph, metapath, damping=0.5):
```
dhimmel: I like the new more efficient implementation (avoiding dense matrices which could cause memory problems). However, I think we want an option to yield zero-DWPC rows.

dhimmel:
Do we need to track all of these files? I think it's too many to merge. What do we want to do?

zietzm:
Outdated by the gamma hurdle model and newer pipeline code

PR #122, opened 2018-06-21
Add new permutations

zietzm:


dhimmel:
GitHub says infinite files changed haha. We may need to rethink tracking every file, and instead switching to a compressed archive approach.

zietzm:
Got the following error when trying to checkout this branch from a different device. 
We must reduce the number of files.

```
Error downloading object: data/hetionet-v1.0.hetmat/permutations/008.hetmat/edges/CtD.sparse.npz (e183a1a): Smudge error: Error downloading data/hetionet-v1.0.hetmat/permutations/008.hetmat/edges/CtD.sparse.npz (e183a1aa8e364444f92169df3dd3f183d3691786c1fc95c77b879fc2d9f13621): batch response: Rate limit exceeded: https://github.com/zietzm/hetmech.git/info/lfs/objects/batch

Errors logged to /home/michael/Documents/hetmech/.git/lfs/logs/20180621T130633.815719752.log
Use `git lfs logs last` to view the log.
error: external filter git-lfs smudge -- %f failed 2
error: external filter git-lfs smudge -- %f failed
fatal: data/hetionet-v1.0.hetmat/permutations/008.hetmat/edges/CtD.sparse.npz: smudge filter lfs failed
```

dhimmel:
> We must reduce the number of files.

Yeah I guess it was a bad idea to track all the permutation files rather than archiving them as one big file. I think we can still track a few permutations in expanded form, but that we should create archives (i.e. booster packs) with the full number of permutations.

dhimmel:
Will close this until we get archives enabled.

PR #124, opened 2018-07-11
ENH: hetmat archiving functionality

dhimmel:


hetmech/hetmat/tests/test_hetmat.py
```diff
@@ -5,10 +5,10 @@
 from hetmech.tests.hetnets import get_graph
 
 
-def test_disease_gene_example_converstion_to_hetmet(tmpdir):
+def test_disease_gene_example_conversion_to_hetmat(tmpdir):
     """
-    Test the PC & DWWC computations in Figure 2D of Himmelstein & Baranzini
-    (2015) PLOS Comp Bio. https://doi.org/10.1371/journal.pcbi.1004259.g002
+    Test converting the hetmat from Figure 2C of https://doi.org/crz8 into a
+    hetmat.
```
dhimmel: These changes are not related to this PR, but I noticed some non-functional issues with this test.

dhimmel:
My commit message from https://github.com/greenelab/hetmech/pull/124/commits/a505a63c86de5651498f8d5963237a0fed60b02f:

> Not sure why data/hetionet-v1.0.hetmat/metagraph.json changed order. This is upsetting as I thought this sort of arbitrary re-ordering was a thing of the past since 3.6 with order-preserving dicts.

It looks like the `kind_to_abbrev` order [from upstream](https://github.com/dhimmel/hetionet/blob/9f214ab7cb6b59fde5fd9530e379d2d71cee4f43/hetnet/json/hetionet-v1.0-metagraph.json#L161-L189) is now preserved, so I think this change is OKAY.

hetmech/hetmat/archive.py
```diff
@@ -0,0 +1,77 @@
...
+def load_archive(archive_path, destination_dir):
+    """
+    Extract the paths from an archive into the specified hetmat directory.
+    """
+    is_url = isinstance(archive_path, str) and re.match('^(http|ftp)s?://', archive_path)
+    if is_url:
+        url = archive_path
+        filename = os.path.basename(url)
+        temp_dir = pathlib.Path(tempfile.mkdtemp())
+        archive_path, _ = urllib.request.urlretrieve(url, temp_dir.joinpath(filename))
```
zietzm: Why generate a directory and name files, when urllib would generate and return a filename anyway?
According to the [urllib docs](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve):
> The second argument, if present, specifies the file location to copy to (if absent, the location will be a tempfile with a generated name).

Seems like we could just do
```python
if is_url:
    archive_path, _ = urllib.request.urlretrieve(archive_path)
with zipfile.ZipFile(archive_path, mode='r') as zip_file:
    zip_file.extractall(destination_dir)
if is_url:
    urllib.request.urlcleanup() 
    # Or, alternatively
    # os.remove(archive_path)
```
The [`urllib.request.urlcleanup()`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlcleanup) automatically removes all the temporary files generated in retrieving a url.

hetmech/hetmat/tests/test_archive.py
```diff
@@ -0,0 +1,42 @@
...
+        'edges/DlT.sparse.npz',
+        'edges/GaD.sparse.npz',
+        'edges/GeT.sparse.npz',
+        'edges/GiG.sparse.npz',
+        'metagraph.json',
+        'nodes/Disease.tsv',
+        'nodes/Gene.tsv',
+        'nodes/Tissue.tsv',
+    ]
+    assert name_list == expected
```
zietzm: Are the names in a zip file always ordered the same way? If not, should we use sets or sorted lists to be safe?

hetmech/hetmat/archive.py
dhimmel: Genius + the elegance of `urlcleanup`... remarkable!

hetmech/hetmat/tests/test_archive.py
dhimmel: Yes we want to test that the files are sorted in a deterministic manner

PR #125, opened 2018-07-12
Bulk computation pipeline

zietzm:
This PR includes efficiency upgrades to `degree_group.py` and a new file including useful functions for the bulk computation of DWPC and DGP.

### DIFF
--- a/hetmech/bulk_computation.py
+++ b/hetmech/bulk_computation.py
@@ -10,32 +10,6 @@ import hetmech.degree_weight
 import hetmech.hetmat
 
 
-def compute_save_dwpc(graph, metapath, damping=0.5, dense_threshold=1, dtype='float64', approx_ok=True,
-                      return_result=True):
-    """
-    Returns and saves a DWPC matrix. If already computed and return_result=True, will return saved result.
-    Automatically ignores inverses, corrupt files, and previously existing files.
-    """
-    metapath = graph.metagraph.get_metapath(metapath)
-    for inverse in True, False:
-        mp = metapath.inverse if inverse else metapath
-        try:
-            return graph.read_path_counts(mp, 'dwpc', damping)
-        except FileNotFoundError:
-            continue
-        except Exception as error:
-            logging.info(
-                f"{metapath}: Path count file read error - {error}"
-            )
-            # Catch all other file issues
-            os.remove(graph.get_path_counts_path(mp, 'dwpc', damping, 'sparse.npz'))
-    row, col, dwpc_matrix = hetmech.degree_weight.dwpc(graph, metapath, damping=damping,
-                                                       dense_threshold=dense_threshold, dtype=dtype)
-    path = graph.get_path_counts_path(metapath, 'dwpc', damping, None)
-    hetmech.hetmat.save_matrix(dwpc_matrix, path)
-    return row, col, dwpc_matrix
-
-
 def compute_save_dgp(hetmat, metapath, damping=0.5, compression='gzip', delete_intermediates=True):
     """
     Compute summary file of combined degree-grouped permutations (DGP). Aggregates across permutations,
--- a/hetmech/bulk_computation.py
+++ b/hetmech/bulk_computation.py
@@ -1,0 +1,86 @@
+import os
+
+import pandas
+import scipy.special
+import scipy.stats
+
+import hetmech.degree_group
+import hetmech.degree_weight
+import hetmech.hetmat
+
+
+def compute_save_dwpc(graph, metapath, damping=0.5, dense_threshold=1, dtype='float64', approx_ok=True,
+                      return_result=True):
+    """
+    Returns and saves a DWPC matrix. If already computed and return_result=True, will return saved result.
+    Automatically ignores inverses, corrupt files, and previously existing files.
+    """
+    metapath = graph.metagraph.get_metapath(metapath)
+    for inverse in (True, False):
+        if inverse:
+            mp = metapath.inverse
+        try:
+            return graph.read_path_counts(mp, 'dwpc', damping)
+        except FileNotFoundError:
+            continue
+        except:
+            os.remove(graph.get_path_counts_path(mp, 'dwpc', damping, 'sparse.npz'))
+    row, col, dwpc_matrix = hetmech.degree_weight.dwpc(graph, metapath, damping=damping,
+                                                       dense_threshold=dense_threshold, dtype=dtype)
+    path = graph.get_path_counts_path(metapath, 'dwpc', damping, None)
+    hetmech.hetmat.save_matrix(dwpc_matrix, path)
+    return row, col, dwpc_matrix
+
+
+def compute_save_dgp(hetmat, metapath, damping=0.5, compression='gzip'):
+    for mp in (metapath.inverse, metapath):
+        combined_path = hetmat.directory.joinpath('adjusted-path-counts', 'dwpc-0.5',
+                                                  'degree-grouped-permutations',
+                                                  f'{mp}.tsv')
+        if combined_path.exists():
+            return
+
+    _, _, matrix = hetmat.read_path_counts(metapath, 'dwpc', damping)
+    matrix_mean = matrix.mean()
+
+    for name, permat in hetmat.permutations.items():
+        path = permat.directory.joinpath('degree-grouped-path-counts', 'dwpc-0.5', f'{metapath}.tsv')
+        if path.exists():
+            pass
+        else:
+            degree_grouped_df = hetmech.degree_group.single_permutation_degree_group(
+                permat, metapath, dwpc_mean=matrix_mean, damping=damping)
+            path.parent.mkdir(parents=True, exist_ok=True)
+            degree_grouped_df.to_csv(path, sep='\t')
+
+    degree_stats_df = hetmech.degree_group.summarize_degree_grouped_permutations(hetmat, metapath, damping=damping,
+                                                                                 delete_intermediates=True)
+    combined_path.parent.mkdir(parents=True, exist_ok=True)
+    degree_stats_df.to_csv(combined_path, sep='\t', compression=compression)
+
+
+def combine_dwpc_dgp(graph, metapath, damping):
+    """
+    Combine DWPC information with degree-grouped permutation summary metrics.
+    Save resulting tables as one-per-metapath, compressed .tsv files.
+    """
+    stats_path = graph.directory.joinpath('adjusted-path-counts', f'dwpc-{float(damping)}',
+                                          'degree-grouped-permutations', f'{metapath}.tsv')
+    degree_stats_df = pandas.read_table(stats_path, compression='gzip')
+
+    dwpc_row_generator = hetmech.degree_group.dwpc_to_degrees(graph, metapath)
+    dwpc_df = pandas.DataFrame(dwpc_row_generator)
+    df = (
+        dwpc_df
+        .merge(degree_stats_df, on=['source_degree', 'target_degree'])
+        .drop(columns=['source_degree', 'target_degree'])
+    )
+    df['r-dwpc'] = df['dwpc'] - df['mean']
+    df['z-dwpc'] = df['r-dwpc'] / df['sd']
+    df['normal-p'] = 1 - scipy.stats.norm.cdf(df['z-dwpc'])
+
+    df['mean-nz'] = df['mean'] * df['n'] / df['nnz']
+    df['beta'] = df['mean-nz'] / df['sd'] ** 2
+    df['alpha'] = df['mean-nz'] * df['beta']
+    df['p-value'] = (df['nnz'] / df['n']) * scipy.special.gammaincc(df['alpha'], df['beta'] * df['dwpc'])
+    return df
--- a/hetmech/bulk_computation.py
+++ b/hetmech/bulk_computation.py
@@ -1,3 +1,4 @@
+import logging
 import os
 
 import pandas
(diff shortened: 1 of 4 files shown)


hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def compute_save_dwpc(graph, metapath, damping=0.5, dense_threshold=1, dtype='float64', approx_ok=True,
+                      return_result=True):
+    """
+    Returns and saves a DWPC matrix. If already computed and return_result=True, will return saved result.
+    Automatically ignores inverses, corrupt files, and previously existing files.
+    """
+    metapath = graph.metagraph.get_metapath(metapath)
+    for inverse in (True, False):
```
dhimmel: parentheses can safely be removed

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+        if inverse:
+            mp = metapath.inverse
```
dhimmel: seems problematic... does `mp` get set if not inverse? Should it be:

```
mp = metapath.inverse if inverse else metapath
```

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            return graph.read_path_counts(mp, 'dwpc', damping)
+        except FileNotFoundError:
+            continue
+        except:
```
dhimmel: `except Exception as error` and then log it.

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Returns and saves a DWPC matrix. If already computed and return_result=True, will return saved result.
+    Automatically ignores inverses, corrupt files, and previously existing files.
+    """
+    metapath = graph.metagraph.get_metapath(metapath)
+    for inverse in (True, False):
+        if inverse:
+            mp = metapath.inverse
+        try:
+            return graph.read_path_counts(mp, 'dwpc', damping)
```
dhimmel: Should `read_path_counts` handle reading the inverse?

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+        combined_path = hetmat.directory.joinpath(
+          'adjusted-path-counts', 'dwpc-0.5', 'degree-grouped-permutations', f'{mp}.tsv')
+        if combined_path.exists():
+            return
+
+    _, _, matrix = hetmat.read_path_counts(metapath, 'dwpc', damping)
+    matrix_mean = matrix.mean()
+
+    for name, permat in hetmat.permutations.items():
+        path = permat.directory.joinpath('degree-grouped-path-counts', 'dwpc-0.5', f'{metapath}.tsv')
```
dhimmel: Let's discuss how to fix the w hardcode

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,87 @@
...
+    dwpc_df = pandas.DataFrame(dwpc_row_generator)
+    df = (
+        dwpc_df
+        .merge(degree_stats_df, on=['source_degree', 'target_degree'])
+        .drop(columns=['source_degree', 'target_degree'])
+    )
+    df['mean-nz'] = df['mean'] * df['n'] / df['nnz']
+    df['beta'] = df['mean-nz'] / df['sd'] ** 2
+    df['alpha'] = df['mean-nz'] * df['beta']
+    df['p-value'] = (df['nnz'] / df['n']) * scipy.special.gammaincc(df['alpha'], df['beta'] * df['dwpc'])
```
dhimmel: seems like the first parentheses are not needed. OKAY to keep them if you think it makes it more readable.

hetmech/degree_group.py
```diff
@@ -72,13 +72,11 @@ def compute_summary_metrics(df):
 
 def dwpc_to_degrees(graph, metapath, damping=0.5):
```
dhimmel: What do the changes in this fxn do?

hetmech/bulk_computation.py
zietzm: My thinking is this: while `read_path_counts` can automatically cycle through the inverses and `dwwc` matrices on disk, the problem is that we can't know what path attempted a load which failed for a reason other than `FileNotFoundError`. These are the cases where we want to delete the corrupt file and replace it with a re-computation.

zietzm: One thing we could do is have `get_path_counts_path` return the path to the file that _would_ be loaded using `read_path_counts`. This way, we could pinpoint where the file read is failing.

hetmech/degree_group.py
zietzm: Two things:
1. Fix the error whereby the `path_counts` column shows `dwpc` values instead.
2. Return only rows having `dwpc` > 0. This was necessary because `G---G` metapaths were exceeding the memory on my computer when bringing these into pandas dataframes.

hetmech/hetmat/__init__.py
```diff
@@ -142,7 +144,12 @@ def read_first_matrix(specs):
             continue
         transpose = spec.get('transpose', False)
         file_format = spec.get('file_format', 'infer')
-        matrix = read_matrix(path, file_format=file_format)
+        try:
+            matrix = read_matrix(path, file_format=file_format)
+        except Exception as error:
+            logging.info(f'{error} reading matrix at {path}')
```
dhimmel: Let's log this with severity warning not info. I would do:

```
logging.warning(f'Error reading matrix at {path}:\n{error}')
```

hetmech/bulk_computation.py
dhimmel: We should probably make some functions to receive these paths in the HetMat class?

hetmech/hetmat/__init__.py
```diff
@@ -142,7 +144,12 @@ def read_first_matrix(specs):
             continue
         transpose = spec.get('transpose', False)
         file_format = spec.get('file_format', 'infer')
-        matrix = read_matrix(path, file_format=file_format)
+        try:
+            matrix = read_matrix(path, file_format=file_format)
+        except Exception as error:
+            logging.warning(f'Error reading matrix at {path}:\n{error}')
+            if delete_failures:
+                os.remove(path)
```
dhimmel: We need a `continue` here. Also let's log the file deletion if it occurs.

dhimmel: No need to import os because path is a pathlib.Path, so `path.unlink()` will work.

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def compute_save_dgp(hetmat, metapath, damping=0.5, compression='gzip', delete_intermediates=True):
+    """
+    Compute summary file of combined degree-grouped permutations (DGP). Aggregates across permutations,
+    deleting intermediates if delete_intermediates=True. Saves resulting files as compressed .tsv files
+    using compression method given by compression. Does not recompute previously saved files.
+    """
+    for mp in (metapath.inverse, metapath):
+        combined_path = hetmat.get_summary_degree_group_path(metapath, 'dwpc', damping)
```
dhimmel: Not ideal to use compression but not use compression extension.

We could create a compression argument for `get_summary_degree_group_path` that would get the extension using a dict like [here](https://github.com/pandas-dev/pandas/blob/6008d75ac00aa7c3052b74bf4f6eb73beeaf56a6/pandas/io/common.py#L253-L258)

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,58 @@
...
+import hetmech.degree_weight
+import hetmech.hetmat
+
+
+def compute_save_dgp(hetmat, metapath, damping=0.5, compression='gzip', delete_intermediates=True):
+    """
+    Compute summary file of combined degree-grouped permutations (DGP). Aggregates across permutations,
+    deleting intermediates if delete_intermediates=True. Saves resulting files as compressed .tsv files
+    using compression method given by compression. Does not recompute previously saved files.
+    """
```
dhimmel: Currently the outer loop is metapath and the inner loop is permutation. For efficiency and caching-purposes, we would like the outer loop to be permutation and have running-DGP metrics updated as metapaths get computed.

This can be saved for a future PR.

hetmech/bulk_computation.py
```diff
@@ -0,0 +1,58 @@
...
+
+    _, _, matrix = hetmat.read_path_counts(metapath, 'dwpc', damping)
+    matrix_mean = matrix.mean()
+
+    for name, permat in hetmat.permutations.items():
+        path = permat.get_degree_group_path(metapath, 'dwpc', damping)
+        if not path.exists():
+            degree_grouped_df = hetmech.degree_group.single_permutation_degree_group(
+                permat, metapath, dwpc_mean=matrix_mean, damping=damping)
+            path.parent.mkdir(parents=True, exist_ok=True)
```
dhimmel: I wonder whether we should do directory creation stuff in the path getters? Perhaps an optional `initialize` or `mkdir` arg?

hetmech/degree_group.py
dhimmel: Can we add a docstring:

> Yield a description of each cell in a DWPC matrix adding source and target node degree info as well as the corresponding path count.

hetmech/degree_group.py
```diff
@@ -88,23 +86,26 @@ def dwpc_to_degrees(graph, metapath, damping=0.5):
...
             'target_id': col_names[col_ind],
             'source_degree': source_degrees[row_ind],
             'target_degree': target_degrees[col_ind],
-            'dwpc': dwpc_matrix[row_ind, col_ind],
-            'path-count': path_count[row_ind, col_ind],
+            'dwpc': dwpc_matrix.data[ind],
+            'path-count': path_count.data[ind],
         }
         yield row
         continue
```
dhimmel: This continue seems peculiar

hetmech/degree_group.py
dhimmel: I think we want the make skipping zero-PC cells an option but not necessarily the default or only option.

hetmech/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    degree_stats_df = pandas.read_table(stats_path, compression='gzip')
+
+    dwpc_row_generator = hetmech.degree_group.dwpc_to_degrees(graph, metapath)
+    dwpc_df = pandas.DataFrame(dwpc_row_generator)
+    df = (
+        dwpc_df
+        .merge(degree_stats_df, on=['source_degree', 'target_degree'])
+        .drop(columns=['source_degree', 'target_degree'])
+    )
+    df['mean-nz'] = df['mean'] * df['n'] / df['nnz']
```
dhimmel: Let's use underscores not `-` in column names to allow convenient reference.

hetmech/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+        hetmat, metapath, damping=damping, delete_intermediates=delete_intermediates)
+    combined_path.parent.mkdir(parents=True, exist_ok=True)
+    degree_stats_df.to_csv(combined_path, sep='\t', compression=compression)
+
+
+def combine_dwpc_dgp(graph, metapath, damping):
+    """
+    Combine DWPC information with degree-grouped permutation summary metrics.
+    Save resulting tables as one-per-metapath, compressed .tsv files that include
+    gamma-hurdle significance estimates.
```
dhimmel: Function doesn't save TSV like the docstring indicates.

hetmech/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    dwpc_df = pandas.DataFrame(dwpc_row_generator)
+    df = (
+        dwpc_df
+        .merge(degree_stats_df, on=['source_degree', 'target_degree'])
+        .drop(columns=['source_degree', 'target_degree'])
+    )
+    df['mean-nz'] = df['mean'] * df['n'] / df['nnz']
+    df['beta'] = df['mean-nz'] / df['sd'] ** 2
+    df['alpha'] = df['mean-nz'] * df['beta']
+    df['p-value'] = df['nnz'] / df['n'] * scipy.special.gammaincc(df['alpha'], df['beta'] * df['dwpc'])
```
dhimmel: Needs fix as @zietzm has prototyped.

hetmech/degree_group.py
```diff
@@ -126,24 +136,25 @@ def single_permutation_degree_group(permuted_hetmat, metapath, dwpc_mean, dampin
...
-        path = permat.directory.joinpath('degree-grouped-path-counts', f'dwpc-{float(damping)}',
-                                         f'{metapath}.tsv')
+        path = permat.get_degree_group_path(metapath, 'dwpc', damping)
         df = (
             pandas.read_table(path)
             .set_index(['source_degree', 'target_degree'])
             .assign(n_perms=1)
         )
+        if delete_intermediates:
+            os.remove(path)
```
dhimmel: `path.unlink`

hetmech/hetmat/__init__.py
```diff
@@ -324,6 +332,32 @@ def get_path_counts_path(self, metapath, metric, damping, file_format):
...
+            'zip': '.zip',
+            'xz': '.xz',
+            None: '',
+        }
+        compr = compression_extension[compression]
+        path = self.directory.joinpath('adjusted-path-counts', f'{metric}-{damping}',
+                                       'degree-grouped-permutations', f'{metapath}.tsv{compr}')
+        return path
+
+    def get_metapath_summary_path(self, metapath, metric, damping):
```
dhimmel: Let's add compression here as well.

PR #128, opened 2018-07-17
create_archive: enable splitting zip output

dhimmel:
```
Co-authored-by: Michael Zietz <michael.zietz@gmail.com>
```


hetmech/hetmat/archive.py
```diff
@@ -39,23 +41,76 @@ def create_archive_by_globs(
...
+
+
+def get_archive_info_df(zip_paths):
+    """
+    Return member file info for a list of zip archives.
+    """
+    fields = [
+        'filename',
+        'file_size',
+        'compress_type',
```
dhimmel: Would be nice to convert `compress_type` to a description rather than int as per https://github.com/python/cpython/blob/6020d98beaf6b9945beb7560f9a1df361e0ba4dd/Lib/zipfile.py#L620-L638

PR #131, opened 2018-07-20
Pipeline: use grouped writer for adjusted-DWPCs

dhimmel:
Enables using pandas.to_csv on generators of rows that are too big for memory.


dhimmel:
We should eventually move this code to a module

PR #133, opened 2018-07-25
Bulk computation for all metapaths <= length 3

dhimmel:
Completed notebook for all 785 metapaths up to length 3 starting on Disease.


dhimmel:
> Have we decided where it would be best to store archives?

Not yet!

PR #135, opened 2018-08-15
[WIP] Gamma hurdle goodness-of-fit

zietzm:


dhimmel:
I took a look at [`660c0bb`](https://github.com/greenelab/hetmech/blob/660c0bb52c0d1a6d7c9b35315e10009dfa968359/explore/gamma-fit.ipynb). While the fits look good to my :eye:, the KS p-values would tend to disagree. As we discussed in person, perhaps there is a missing third parameter in our gamma distribution formulation.

PR #136, opened 2018-08-16
Distribution of p-values hetionet vs permutation

zietzm:


dhimmel:
Regarding bin_counts.tsv, can we combine all into a single TSV by just adding another column for metapath. Can we save raw counts rather than percentages. Also set float_format (see the floating point issue in some of the percentiles.)

PR #137, opened 2018-08-21
[WIP]: Empirical p-value NB

zietzm:
image


zietzm:
@dhimmel I randomly sampled 5000 rows (so that the computation won't take all day) and did the simplest empirical p-value calculation, as we discussed. Here is that distribution

image

I will try to overlay this with a permutation asap

dhimmel:
> I randomly sampled 5000 rows

Good idea.

Did you change the p-value calculation to be :skull: simple... i.e. no multiplication and just a single inequality comparison?

PR #138, opened 2018-08-21
[WIP]: Explore p-values for simulated DWPC values

zietzm:
image


zietzm:
@dhimmel I think the issue was that I was inverting number nonzero and number zero when simulating the data. Now using the normal method I get the following simulations:
image
The second one takes the actual sets of gamma hurdle distribution parameters (alpha, beta, nnz/n) and generates random draws from those distributions. Therefore, this one is a composite of several distributions, as you described.
image
What do you think about the second one having a left side above one?
This is it with more bins:
image

cgreene:
🎉 !

zietzm:
@dhimmel You were exactly right about averaging the PDFs! Here are two plots doing this correctly, both showing enrichment.

image

image

PR #140, opened 2018-08-23
DWPCs & null DWPCs for all metapaths up to length 3

dhimmel:


dhimmel:
In https://github.com/greenelab/hetmech/pull/140/commits/2387c1c25ab295d282b1d20e20f7e30194cd6de5, we ran the pipeline for two testing metapaths: DrD and SpDpS. In https://github.com/greenelab/hetmech/pull/140/commits/92f40fe5c9e4f055d861587dd30ad1455aecba8a, we update the archive files. Note the file sizes and checksums change for the DGP files, which is expected because the DGP files only contain 99 permutations due to the process dying mid-computation. However, the DWPC checksums are the same.

dhimmel:
In https://github.com/greenelab/hetmech/issues/141 I describe the memory leak in more detail. In the meantime I will merge this despite the incomplete DGP computation.

I'm still working on uploading the PC / DWPC matrices generated here to Zenodo. The 50GB policy applies to an entire record not just a single file, so I've contacted them for an exception. Fingers crossed.

Will merge now and subsequent updates in future PRs.

Issue #145, opened 2018-10-26
Fix error when importing hetmech.matrix


zietzm:
@gwaygenomics made the following observation:

Importing `hetmech.matrix` without previously importing `hetmech.hetmat` gives an `ImportError`

In the following image, I ran the first cell, then restarted the kernel and ran only the second cell. As can be seen below, importing `hetmech.matrix` alone or before `hetmech.hetmat` gives this error.

![image](https://user-images.githubusercontent.com/7230303/47578709-159f3d00-d918-11e8-9d2a-058bb3e0efc8.png)

gwaygenomics:
I am getting a different, but related error now when using `hetmech` in a conda environment with:

```sh

PR #146, opened 2018-10-26
Fix import errors

zietzm:
Addresses #145 

While the errors were raised by `import hetmech.matrix`, the real issue was in the circularity of imports between `hetmech.matrix` and `hetmech.degree_weight`. Removing the individual imports and instead importing the full files [has been suggested](https://stackoverflow.com/questions/22187279/python-circular-importing) as a way to remedy any issues with circular importing. 

I have gone through `hetmech.degree_weight` and updated all `matrix` functions. Interestingly, I [found an instance](https://github.com/greenelab/hetmech/compare/master...zietzm:imports?expand=1#diff-dcb7103539fb7297ec924e8232f7736eR147) of `hetmech.matrix.metaedge_to_adjacency_matrix`, which may have actually previously been a bug in the code, as `hetmech.matrix` was not explicitly imported in the past.

### DIFF
--- a/hetmech/degree_weight.py
+++ b/hetmech/degree_weight.py
@@ -12,11 +12,7 @@ from hetio.matrix import (
 
 import hetmech.hetmat
 from hetmech.hetmat.caching import path_count_cache
-from hetmech.matrix import (
-    copy_array,
-    metaedge_to_adjacency_matrix,
-    normalize,
-)
+import hetmech.matrix
 
 
 def _category_to_function(category, dwwc_method):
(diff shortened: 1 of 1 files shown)


dhimmel:
> Interestingly, I found an instance of hetmech.matrix.metaedge_to_adjacency_matrix, which may have actually previously been a bug in the code, as hetmech.matrix was not explicitly imported in the past.

Odd that our test for dwwc_recursive did not fail:

https://github.com/greenelab/hetmech/blob/b4aca774a2e9b41e6c448535d3b1cc2d9b3f85f4/hetmech/tests/test_degree_weight.py#L26

Perhaps it's because hetmech.matrix got loaded from elsewhere...

I am on board with the changes, but want to take a bit more time to research why the workaround works and how it is different.

PR #147, opened 2018-10-30
add find_packages()

gwaygenomics:
refs https://github.com/greenelab/hetmech/issues/145#issuecomment-434412918

This enables the conda environment to see the `hetmat` folder during install

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,12 @@
 from setuptools import setup
+from setuptools import find_packages
 
 setup(name='hetmech',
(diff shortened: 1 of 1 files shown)


gwaygenomics:
looks like this passed checks - what do you think @dhimmel @zietzm ?

dhimmel:
Okay looks like this solves #145 / https://stackoverflow.com/a/50120983. Thanks @gwaygenomics. I gotta use find_packages more cause these errors are REALLY ANNOYING

PR #149, opened 2018-12-11
Relocate package to hetio/hetmatpy

dhimmel:
Refs https://github.com/hetio/hetmatpy/issues/1


.travis.yml
```diff
@@ -11,10 +11,7 @@ before_install:
 install:
   - conda env create --quiet --file environment.yml
   - conda activate hetmech
-  - pip install --editable .
   - conda list
 script:
-  - pytest hetmech
```
zietzm: Do we need to keep the Travis functionality at all? It seems that the notebook flake8 checking is spotty at best.

dhimmel: The ability to successfully install an environment specification is a feat in and of itself (:  

We could disable in future by deleting `.travis.yml` or just flipping the travis toggle.

PR #151, opened 2019-01-24
Example notebook with consolidated path scores across metapaths

dhimmel:
Rotation student @ben-heil has been working on combining scores between metapaths. He's created an example notebook titled `9.global-path-weighting.ipynb` which provides each path between two nodes with a score in a unified table.


environment.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
   - conda-forge::tqdm=4.24.0
   - conda-forge::xarray=0.10.8
   - conda-forge::xlrd=1.1.0
+  - conda-forge::psycopg2-2.7.5
```
dhimmel: separate version with `=` not `-`

dhimmel:
Would be helpful to get a column that shows -log10(p-value) for the metapath.

ben-heil:
Fixed the environment file and added the metapath importance column

dhimmel:
Importance plot is nice. Another way we could show the importance plot: cumulative percent of DWPC. This would allow us to say "the top 10 paths account for 95% of the DWPC"

dhimmel:
Why is `PERCENT_OF_DWPC` ALL_CAPS? Is that something we did in hetio?

ben-heil:
> Importance plot is nice. Another way we could show the importance plot: cumulative percent of DWPC. This would allow us to say "the top 10 paths account for 95% of the DWPC"

I'm not sure what you mean by this. Would the cumulative percent be the sum of the percents of the DWPCs for the top n paths divided by the total number of metapaths returned * 100?

> Why is PERCENT_OF_DWPC ALL_CAPS? Is that something we did in hetio?
Yeah, that's the way it returns from the cypher query from hetio. I don't remember why I wrote it that way though, it could have been that I was trigger happy with the caps lock because I was writing Cypher. Should I change it?

dhimmel:
> I'm not sure what you mean by this. Would the cumulative percent be the sum of the percents of the DWPCs for the top n paths divided by the total number of metapaths returned * 100?

I overlooked that this chart combined importance for paths of many metapaths, so my comment is less directly applicable. 

> Should I change it?

Let's fix the capitalization to be consistent and generally lower case for variable names. But that is independent from this PR.

PR #152, opened 2019-01-28
Changed the case of percent_of_DWPC to match the change in hetio

ben-heil:
Keep the notebook from breaking due to [this change](https://github.com/hetio/hetio/pull/34)


dhimmel:
Don't we also have to update

https://github.com/greenelab/hetmech/blob/ebd439a10bc542779389f8c902a6905aaf89d88b/environment.yml#L24

ben-heil:
Good catch, I forgot that I had my hetmech environment to automatically use updated versions of hetio

PR #155, opened 2019-02-08
Added notebook to compare compound rankings 

ben-heil:
Found that the correlation between unsupervised rankings (the sum of the log p_values for all metapaths) are weakly positively correlated with the rankings from rephetio


ben-heil:
I still haven't been able to find what is causing compounds highly ranked in the supervised method to perform poorly by the unsupervised method and vice versa. I have found that both methods perform well in ranking drugs known to treat a disease very highly.

Here's a plot comparing unsupervised and supervised rankings for drugs known to treat a disease (status 1).
image

Here's a pair of histograms (technically frequency polygons) comparing where both methods rank the known drugs for all diseases in the dataset
image

PR #157, opened 2019-02-15
Examine parameter estimates

zietzm:
Explanation from notebook pasted below for future reference:

# Gamma-hurdle parameter estimation

Until present, we have been estimating the parameters of the gamma-hurdle distribution as follows. Suppose we observe , draws from a gamma-hurdle distribution. 

&space;0\}" title="x_{nz} \equiv \{x_i | x_i \in x, x_i > 0\}" />

}{<x^2_{nz}>&space;-&space;<x_{nz}>^2}" title="\hat\beta = \frac{<x_{nz}>}{<x^2_{nz}> - <x_{nz}>^2}" />

^2}{<x^2_{nz}>&space;-&space;<x_{nz}>^2}" title="\hat\alpha = \frac{<x_{nz}>^2}{<x^2_{nz}> - <x_{nz}>^2}" />

This method is very computationally favorable, as it requires only the sum, sum of squares, `n`, and `nnz` to estimate all three parameters. These estimates were determined by matching mean and variance to those of a gamma distribution given certain parameters, rather than by a more standard method.

A standard way to estimate the parameters of statistical distributions given observations is maximum likelihood estimation (MLE). Using MLE, we obtain the same estimate for  and the same relationship between  and  (that is, " title="\hat\alpha = \hat\beta <x_{nz}>" />), but a different estimate for  itself. The MLE estimates are as follows

\bigg)&space;-&space;<&space;\log(x_{nz})&space;>" title="s \equiv \log \bigg(<x_{nz}>\bigg) - < \log(x_{nz}) >" />

}" title="\hat\beta \approx \frac{3 - s + \sqrt{(s-3)^2 + 24s}}{12s <x_{nz}>}" />

This notebook attempts to determine whether our moment-matching estimates are similar to the MLE estimates.


explore/gamma-hurdle/parameter_estimates.ipynb
```diff
@@ -0,0 +1,169 @@
...
+      "text/plain": [
+       "<Figure size 432x288 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plt.scatter(df['b_mle'], df['b_match'])\n",
```
ben-heil: Do you think the increase in our estimator's variance as beta gets larger will negatively affect our model? Do we have reason to believe that when beta gets into the 20-40 range in the database should we switch to the MLE estimator? Or should they be about the same accuracy-wise?

zietzm: The MLE estimate here is approximate, while the method of moments estimate is exact. I don't think estimation quality (various [advantages](https://en.wikipedia.org/wiki/Method_of_moments_(statistics)#Advantages_and_disadvantages) of MLE over method of moments) isn't a major concern, since the quasi p-values computed are not actual hypothesis tests, just a way to determine which DWPC values are more or less likely to have occurred due to degree alone.

explore/gamma-hurdle/parameter_estimates.ipynb
```diff
@@ -0,0 +1,169 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Gamma-hurdle parameter estimation\n",
```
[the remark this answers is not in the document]
zietzm: Very strange that they are hidden since they are certainly in the file

PR #158, opened 2019-02-25
Comparison of rephetio metapath contributions and unsupervised p-values

ben-heil:
This noteboook is similar to the [one comparing compound rankings](https://github.com/greenelab/hetmech/pull/155), except that it looks at individual metapaths for each compound instead of compounds as a whole. Interestingly, where the other notebook finds a weak but statistically significant correlation, this notebook finds no correlation between the p values for a metapath and the weight the metapath is assigned in a supervised model.

 
image


ben-heil:
I talked with Daniel about the result above and he said that [the file that I was using](https://github.com/dhimmel/learn/blob/master/prediction/predictions/metapath-contribution.tsv.bz2) didn't contain the actual model weights but the weights for the model normalized so that each compound/disease pair's positive weights would add to one. It is reasonable that there wouldn't be a correlation, because a metapath's rank within compound/disease pair should only correlate weakly with the global p-value.

Upon using the [raw positive weights](https://github.com/dhimmel/learn/blob/master/prediction/predictions/term-contribution.tsv.bz2), we get the cool results below. I'm not sure what causes the stratification (maybe @dhimmel does), but it's encouraging that the stronger metapath weights have stronger p-values.

image
image

dhimmel:
> I'm not sure what causes the stratification (maybe @dhimmel does), but it's encouraging that the stronger metapath weights have stronger p-values.

Color by metapath and thou shall see.

ben-heil:
The stratifications seem to be strongly related to individual metapaths

image

image

ben-heil:
Good call, the legends were already ordered by y-value (I posted the wrong image), but the standard error curves are now colored according to their metapaths.

image

PR #159, opened 2019-03-08
Upgraded logic to switch row elements if the disease is in the source column

ben-heil:


11.compound_ranking_comparison.ipynb
```diff
@@ -50,32 +50,30 @@
...
-    "             top_ids.target_id = outer_pc.source_id)\n",
-    "    ORDER BY outer_pc.p_value;\n",
+    "    query = f'''SELECT dwpc, p_value, metapath_id, source_node.name AS source_name, \n",
+    "                       target_node.name AS target_name, source_node.identifier AS source_identifier \n",
+    " FROM dj_hetmech_app_pathcount pc \n",
+    " JOIN dj_hetmech_app_node source_node \n",
+    "     ON pc.source_id = source_node.id \n",
+    " JOIN dj_hetmech_app_node target_node \n",
+    "     ON pc.target_id = target_node.id     \n",
+    " WHERE target_node.identifier='{disease_id}' OR source_node.identifier='{disease_id}'\n",
```
dhimmel: Do we also need to check node type is Disease?

11.compound_ranking_comparison.ipynb
```diff
@@ -50,32 +50,30 @@
...
-    "             ON pc.target_id = n2.id     \n",
-    "         WHERE join_node.identifier='{disease_id}' \n",
-    "         ORDER BY pc.p_value) AS top_ids \n",
-    "    JOIN dj_hetmech_app_pathcount outer_pc \n",
-    "         ON (top_ids.source_id = outer_pc.source_id AND \n",
-    "             top_ids.target_id = outer_pc.target_id) OR \n",
-    "             (top_ids.source_id = outer_pc.target_id AND \n",
-    "             top_ids.target_id = outer_pc.source_id)\n",
-    "    ORDER BY outer_pc.p_value;\n",
+    "    query = f'''SELECT dwpc, p_value, metapath_id, source_node.name AS source_name, \n",
```
dhimmel: Would be great to improve formatting so the query all has the same starting indentation. Could split query out of fxn.

11.compound_ranking_comparison.ipynb
ben-heil: The identifier provided is a Disease Ontology id, I wouldn't expect them to map to any other identifiers. Are there other data sources that also use the 'DOID:xxxx' format?

ben-heil: I could also add logic to verify that the identifier passed in is a DOID if we plan to use this function somewhere else in the future

dhimmel: In this case, the `DOID:` prefix of disease identifiers should prevent collisions. However, generally you must match to metanode-identifier to ensure uniqueness. Since this function is in a notebook and is not meant to be general, that is OKAY. Perhaps just add a comment that if reusing elsewhere, it may be necessary to add metanode matching.

ben-heil: Sounds good, I added a comment

PR #161, opened 2019-03-19
Notebook looking for clusters of compounds based on metapath p-values

ben-heil:
I made a heatmap and some t-SNE plots to get a feel for the data. The heatmap is super sparse (all the whitespace are NaN values), and there don't seem to be clear subgroups falling out of the t-SNE plot. The t-SNE embedded plots are about what I'd expect, with the metapath with the highest average p_values (CcSEcCtD) appearing as the main axis. 

image

image

image


dhimmel:
Very cool! Is [umap](https://github.com/lmcinnes/umap) better than tsne?

ben-heil:
The UMAP embeddings don't show anything that you can't see in the tSNE embeddings as far as I can see, though they are much more efficient computationally.

Also, the most recent version of numba (0.43.0) [breaks UMAP](https://github.com/lmcinnes/umap/issues/209), so I added 0.42.0 to the environment.yml. The authors of UMAP said that they will likely have a fix some time this week, at which we can remove numba from the environment file and let conda sort out the dependency on its own.

image
image

dhimmel:
holymoly! the treatments are the bird's beak!

Would be cool to see more compound sets superimposed on these plots:

+ whether clinical trial investigates indication
+ whether the compound was ictogenic or anti-ictogenic as [per this](https://github.com/dhimmel/rephetio/blob/gh-pages/epilepsy/data/PK-curation.tsv)
+ by pharmacologic categories [here](https://github.com/dhimmel/rephetio/blob/f02d44fde7eeef0ffdca0800e0b43c48d800c86d/epilepsy/data/compounds-categories.tsv). See [this notebook](https://github.com/dhimmel/rephetio/blob/f02d44fde7eeef0ffdca0800e0b43c48d800c86d/epilepsy/3.contribution.ipynb) -- we should be able to plot these categories for more than just the top 100 Rephetio predictions.

Would also be cool to produce an interactive version of this plot with point tooltips showing which compound each point is.

dhimmel:
[Wow](https://nbviewer.jupyter.org/github/greenelab/hetmech/blob/3c528db31c7963a48dfa49c48e9c89e072b3a7e9/explore/compound_clustering.ipynb)

ben-heil:
It's not quite done yet, I realized that the data source I was pulling from had categories for the top 100 compounds instead of all the compounds. I'll have the updated version up in a few minutes (I hope).

ben-heil:
The updated version is up! Plotly makes beautiful visualizations but is fairly unwieldy to work with. Also, we have descriptions for all the compounds which could be added to the tooltips in the visualization. I would have put them in, but plotly doesn't automatically handle things like multiple text fields or word wrapping, so the descriptions would run way off the sides of the visualization.

Additionally, I agree with Daniel that it will be useful to add links to the hetmech search engine to the tooltips once that is possible

PR #168, opened 2019-04-25
Explore setting a metaedge-specific threshold

dhimmel:
Refs https://github.com/greenelab/hetmech-backend/pull/41
Refs https://github.com/greenelab/connectivity-search-manuscript/pull/16


dhimmel:
@zietzm would be great to chat about your ideas on this next time we're both in office. The goal here is to set a different threshold for each metapath, to help import the most meaningful DWPCs given a database size constraint. We might also want to weight path scores by metapath threshold.

